PR #1, opened 2015-03-10
MAID-871 Added tests

lionelkameni:


src/encryption.rs
```diff
@@ -116,5 +116,32 @@ mod test {
     let decrypted_data = decrypt(&encrypted_data[..], &key, &iv).ok().unwrap();
 
     assert!(message.as_bytes() == &decrypted_data[..]);
-  } 
+  }
```
dirvine: This code should be in Self Encryption (lib.rs) and decorate the existing struct. So if you can move it there and update tests etc. then we are good to go.

src/lib.rs
```diff
@@ -87,35 +87,44 @@ pub trait Storage {
...
+
 pub struct SelfEncryptor<'a> {
   storage: &'a mut (Storage + 'a),
   my_datamap: datamap::DataMap,
   chunks: Vec<Chunks>,
   sequencer: Vec<u8>,
-  tempdir : TempDir, 
-  file_size: u64,
+  tempdir : TempDir,
+  pub file_size: u64,
```
dirvine: Is this dangerous Leo as somebody may change it from outside the functions we provide ? Perhaps there is another way to return just a copy of this number ?

src/lib.rs
```diff
@@ -310,16 +337,27 @@ fn test_xor() {
...
 
+
+
 #[test]
 fn check_write() {
-  let mut se = SelfEncryptor::new(datamap::DataMap::None);
+  //struct MyStorage;
+  //let name = vec![0x11];
+  /*impl Storage for MyStorage {
+     fn get(&mut self, name: Vec<u8> ) -> Vec<u8> {name}
```
dirvine: If this is not used we should delete it

src/lib.rs~
```diff
@@ -0,0 +1,449 @@
+/*  Copyright 2014 MaidSafe.net limited
```
dirvine: I think you have pushed temp files. Can you add a .gitignore and also remove these please.

PR #3, opened 2015-03-15
implement DataMap::sort()

maqi:


src/datamap.rs
```diff
@@ -62,5 +72,21 @@ impl DataMap {
...
+            if chunks[j].pre_hash < chunks[cur_min].pre_hash {
+                cur_min = j;
+            }
+            j = j + 1;
+        }
+        chunks.swap(i, cur_min);
+        i = i + 1;
+    }
   }
+}
```
dirvine: I think the sort trait is implemented for this type, can you please check.

maqi: I think the previous implementation of sort trait is incorrect and won't work as expected.
Actually it is get_sorted_chunks needs to be implemented ( and got implemented in this pull request)

PR #4, opened 2015-03-16
Maid 871 encryptor tests

lionelkameni:


src/lib.rs
```diff
@@ -222,13 +222,21 @@ impl<'a> SelfEncryptor<'a> {
...
+    let xor_result = xor(&self.storage.get(name), &kvp.0);
+    return encryption::decrypt(&xor_result, &kvp.2[..], &kvp.1[..]).ok().unwrap();
+  }
+
+    fn encrypt_chunk(&self, chunk_number : u32, content : Vec<u8>)->Vec<u8> {
+    let name = self.my_datamap.get_sorted_chunks()[chunk_number as usize].hash.clone();
+    // [TODO]: work out passing functors properly - 2015-03-02 07:00pm
+    let kvp = &self.get_pad_iv_key(chunk_number);
+    let enc = &encryption::encrypt(&content, &kvp.2[..], &kvp.1[..]).ok().unwrap();
     xor(&enc, &kvp.0)
```
dirvine: This needs to update the datamap to work though. So it should update the post_encryption hash (called just hash). The close method should hash these.  Ideally as we write we need to update the datamap and mark chunks as ToBeHashed and then make sure the current chunk and two preceding ones are all hashed before calling encrypt, or have encrypt check this.

tests/lib.rs
```diff
@@ -67,3 +67,18 @@ fn check_write() {
   assert_eq!(se.len(), 8u64);
   assert_eq!(se.get_storage().get(name),vec![0x11]);
 }
+
+#[test]
+
+fn check_read() {
+  let name = vec![0x11];
+  let mut my_storage = MyStorage{name: vec![0x11]};
+  let mut se = SelfEncryptor::new(&mut my_storage as &mut Storage, datamap::DataMap::None);
```
dirvine: So next test is to create a new encryptor and pass teh correct datamap to the new() call, when the close() if completed of course.

PR #5, opened 2015-03-16
implemented close()

maqi:


dirvine:
Excellent Qi, waiting on travis now

dirvine:
I think close() should return a Result type Qi, then it can be an error of already closed as opposed to empty datamap.

dirvine:
Or perhaps just panic!("closed") in read/write if closed is true

maqi:
The code was "translated" from the current C++ implementation.
However, the current C++ close() is a void function.
but as the API is changed in RUST, I will prefer Result to panic, as raising may means when we translate code currently using close(), caution of panic need to be cared for.

maqi:
But, isn't the original purpose of the definition of DataMap::None is to allow upper layer code a easy match code to handle the wrong situation?
In that case, I think the current return of datamap shall be fine, and actually using Result or Panic may just be an overlap of DataMap::None

PR #8, opened 2015-03-18
extend check_3_min_chunks unit test to cover read(), write() and close()...

maqi:
...; correct the implementation to pass the extended test


dirvine:
Again nice Qi

PR #13, opened 2015-03-18
Various fixes for more idiomatic rust

Ryman:
The tests passed throughout, but I assume the coverage is low, so you may want to take a closer look at b209462, 259773d, 0ee9383, 900a2f3 in particular.

### DIFF
--- a/src/datamap.rs
+++ b/src/datamap.rs
@@ -83,12 +83,8 @@ impl DataMap {
   }
 
   /// iterate through the chunks to figure out the total size, i.e. the file size
-  fn chunks_size(chunks: &Vec<ChunkDetails>)->u64 {
-    let mut size = 0u64;
-    for i in chunks.iter() {
-      size += i.source_size
-    }
-    return size
+  fn chunks_size(chunks: &[ChunkDetails]) -> u64 {
+    chunks.iter().fold(0, |acc, chunk| acc + chunk.source_size)
   }
 
   /// sorting list of chunks using quicksort
--- a/src/datamap.rs
+++ b/src/datamap.rs
@@ -44,7 +44,7 @@ impl DataMap {
   /// returning the list of chunk info if present
   pub fn get_chunks(&self)->Vec<ChunkDetails> {
     match *self {
-      DataMap::Chunks(ref chunks) => return chunks.to_vec(),
+      DataMap::Chunks(ref chunks) => chunks.to_vec(),
         _                         => panic!("no chunks")
     }
   }
--- a/src/datamap.rs
+++ b/src/datamap.rs
@@ -15,7 +15,12 @@ pub struct ChunkDetails {
 
 impl ChunkDetails {
   pub fn new() -> ChunkDetails {
-    ChunkDetails { chunk_num : 0, hash : Vec::new(), pre_hash : Vec::new(), source_size : 0 }
+    ChunkDetails {
+      chunk_num: 0,
+      hash: vec![],
+      pre_hash: vec![],
+      source_size: 0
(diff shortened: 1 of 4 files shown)


dirvine:
Really nice work, amazing to get feedback and valuable input as we rush through learning rust and doing this special ops mission to see how it looks. Fantastic help, please keep in touch as we do have core dev rewards which will become more obvious on launch. It's not about the cash, but it's nice to be recognised, especially in this open source world.   
Will get back if there's anything that needs any comment or clearing up but looks great on first scan.

src/lib.rs
```diff
@@ -42,30 +41,35 @@ extern crate crypto;
...
+pub const MIN_CHUNK_SIZE: u32 = 1024;
+
 /// Helper function to XOR a data with a pad (pad will be rotated to fill the length)
 /// ### Usage
-///    let data = vec![1u8, 2u8, 3u8, 4u8, 5u8];  
-///    let pad = vec![8u8, 7u8];  
+///    let data = vec![1u8, 2u8, 3u8, 4u8, 5u8];
+///    let pad = vec![8u8, 7u8];
 ///    let xor_result = xor(&data, &pad);
-///
```
maqi: this end of line spaces are left over purpose to serve as line break when documentation got generated
same to couple of other places

Ryman: Sorry, that specific one was during a rebase and didn't notice the code sample wasn't wrapped.

You should surround code with \`\`\` if you want it formatted with appropriate spacing. They'll also get tested that they compile and run :) (if you want to just test compile but not run, use \`\`\`{rust, no_run}

maqi: Thanks, will try that.

src/datamap.rs
```diff
@@ -8,103 +8,82 @@ pub struct ChunkDetails {
...
-  pub fn len(&self)->u64 {
-    let size = 0u64;
+  pub fn len(&self) -> u64 {
     match *self {
-      DataMap::Chunks(ref chunks) => DataMap::chunks_size(chunks), 
-        DataMap::Content(ref content) => content.len() as u64, 
-        DataMap::None => 0u64
+      DataMap::Chunks(ref chunks) => DataMap::chunks_size(chunks),
+      DataMap::Content(ref content) => content.len() as u64,
+      DataMap::None => 0
```
benjaminbollen: Is this change okay 0u64 to 0 ?

Ryman: It's redundant as the type is inferred from both the other match arms and the return type.

src/datamap.rs
```diff
@@ -8,103 +8,82 @@ pub struct ChunkDetails {
...
 
   /// iterate through the chunks to figure out the total size, i.e. the file size
-  fn chunks_size(chunks: &Vec<ChunkDetails>)->u64 {
-    let mut size = 0u64;
-    for i in chunks.iter() {
-      size += i.source_size
-    }
-    return size
+  fn chunks_size(chunks: &[ChunkDetails]) -> u64 {
+    chunks.iter().fold(0, |acc, chunk| acc + chunk.source_size)
```
benjaminbollen: Does Rust not need an explicit return statement? sorry for being an absolute Rust novice :)

Ryman: Only for early returns, see here for more detail: http://doc.rust-lang.org/book/functions.html

dirvine: You will see Ben, if for etc. can be used as expressions in rust as well. Pretty powerful for functional type programming. See the rust channel in our slack for some links for c++ programmers -> rust.

PR #15, opened 2015-03-19
Ondisk storage

lionelkameni:
Added disk interface tests

### DIFF
--- a/tests/lib.rs
+++ b/tests/lib.rs
@@ -21,7 +21,14 @@
 
 extern crate self_encryption;
 extern crate rand;
+extern crate rustc_back;
 pub use self_encryption::*;
+use std::path::Path;
+use std::fs::File;
+use rustc_back::tempdir::TempDir as TempDir;
+use std::string::String as String;
+
+
 /// DataMap integratoin tests
 #[test]
 fn data_map_empty(){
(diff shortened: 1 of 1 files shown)


dirvine:
I will leave this open whilst you sort out the logic @lionelkameni  so give me a shout when you fell ready for code review/merge

tests/lib.rs
```diff
@@ -21,7 +21,15 @@
 
 extern crate self_encryption;
 extern crate rand;
+extern crate rustc_back;
```
dirvine: Why this instead of http://web.mit.edu/rust-lang_v1.0al/arch/amd64_ubuntu1404/share/doc/rust/html/std/io/struct.TempDir.html

tests/lib.rs
```diff
@@ -39,48 +47,130 @@ fn data_map_content_only(){
...
 
 #[test]
-fn check_write() {
-  let name = vec![0x11];
-  let mut my_storage = MyStorage{name: vec![0x11]};
-  let mut se = SelfEncryptor::new(&mut my_storage as &mut Storage, datamap::DataMap::None);
-  se.write(&random_string(3), 5u64);
-  assert_eq!(se.len(), 8u64);
-  assert_eq!(se.get_storage().get(name),vec![0x11]);
+fn check_disk_int_5000() {
```
dirvine: Please consider iterating various sizes as opposed to copy paste type code please.

tests/lib.rs
Ryman: Those MIT docs are old, you should use the officially hosted ones, tempdir is now http://doc.rust-lang.org/std/fs/struct.TempDir.html

tests/lib.rs
```diff
@@ -42,46 +50,87 @@ fn random_string(length: u64) -> String {
...
-  let mut my_storage = MyStorage{name: vec![0x11]};
-  let mut se = SelfEncryptor::new(&mut my_storage as &mut Storage, datamap::DataMap::None);
-
-  let the_string = random_string(3);
-  se.write(&the_string, 5u64);
-//    let leng = se.len();
-  let to_be_read = se.read(5u64, 3);
-  assert_eq!(to_be_read, the_string)
+fn check_disk(){
+let mut vec = vec![1, 5000, 10000000];
```
Ryman: This could be extracted into a macro (macro_rules) so three separate tests are created (this is more an issue for benchmarks, but tests run in parallel so best to keep them small)

tests/lib.rs
lionelkameni: Hi Ryman, it says your link is a deprecated library

tests/lib.rs
```diff
@@ -42,46 +50,87 @@ fn random_string(length: u64) -> String {
...
-}*/
-
 pub struct MyStorage {
-    name: Vec<u8>
+  temp_dir : TempDir
+}
+
+impl MyStorage {
+  pub fn new() -> MyStorage {
+    MyStorage { temp_dir: match TempDir::new("encrypt_storage") {
```
Ryman: This formatting is pretty confusing, `tempdir: TempDir::new(..).unwrap()` would suffice for test code. `unwrap` will print the error message on `Result`.

lionelkameni: I seem to recall trying that and getting an error. I will have to try again

tests/lib.rs
Ryman: Yeah, the latest is in fact on cargo https://github.com/rust-lang/tempdir

PR #17, opened 2015-03-20
Maid 922 implement SE integration tests

lionelkameni:
Test check_disk() not passing


examples/SelfEncryptorExample.rs
```diff
@@ -0,0 +1,55 @@
+extern crate self_encryption;
```
dirvine: How does a person use this example ?
Is the naming convention correct for the example? should it not be self_encryptor_example ?

src/lib.rs
```diff
@@ -126,12 +126,74 @@ pub struct SelfEncryptor<'a> {
...
+/// # Example
+///
+/// SelfEncryptor Example
+///
+/// ```rust
+/// extern crate self_encryption
+/// extern crate rand;
+/// use use self_encryption::*;
+///
+/// fn main() {
```
dirvine: Again should this not take some user input to encrypt a file on disk or similar and then also to decrypt it. (see docopts lib for a nice interface).

examples/SelfEncryptorExample.rs
Ryman: This is importing the main lib into the example. Users can run examples with `cargo run --example selfencryptorexample`

src/lib.rs
Ryman: Has `cargo test` been run? This code shouldn't compile (missing semicolon after `extern crate self_encryption`)

src/lib.rs
```diff
@@ -152,13 +214,13 @@ impl<'a> SelfEncryptor<'a> {
...
   }
 
-  /// This is an implementation of the get_storage function from example
+  /// This is an implementation of the get_storage function from example.
   pub fn get_storage(&'a mut self) -> &'a mut Storage { self.storage }
 
-  /// Write method mirrors a posix type write mechanism
-  /// loosly mimics filsystem interface for easy connection to FUSE like
+  /// Write method mirrors a posix type write mechanism.
+  /// It losely mimics a filesystem interface for easy connection to FUSE like
```
Ryman: loosely

tests/lib.rs
```diff
@@ -42,46 +49,80 @@ fn random_string(length: u64) -> String {
...
-  assert_eq!(se.len(), 8u64);
-  assert_eq!(se.get_storage().get(name),vec![0x11]);
-}
 
 #[test]
-fn check_read() {
-  let name = vec![0x11];
-  let mut my_storage = MyStorage{name: vec![0x11]};
-  let mut se = SelfEncryptor::new(&mut my_storage as &mut Storage, datamap::DataMap::None);
+  fn check_disk(){
```
Ryman: Is this not in the other PR?

src/lib.rs
lionelkameni: Hi David,
Hope weekend went well. Productive weekend my end. So I've implemented the
dotcopts interface for user input. I haven't tried compiling yet as I'd
need the office for that but I wondered if I should see with you to just
have a look on logic or post on slack. I notice Rhyman is great at
detecting whether something will work entirely by looking at code. But the
implementation seems good.
Anyways I also made good progress on random write test (who knew
implementing a subtract method in rust was a nightmare), but I am a bit
unsure on how to implement the vector pair in the C++ example. I will ask
on slack if I don't get it sorted tonight. I will have a look at C++
example for benchmark but probably compression should be next on my agenda.
But anyways yea good weekend work so happy.
Best wishes

dirvine: Good Leo nice to see you getting to grips with this, overwhelming perhaps but well worth the effort. 

Pair is simple in rust just use (thing1Type, thing2Type). So you can have Vec<(thinkg1type, thing2type)> or similar. So tuples everywhere more so than c++, but ADT's (enums) in rust also work great if you are only holding one type of many options (even with initialisers). 

Subtract should also be simple, if you post on slack what you are trying I will give you a hand, or post a playpen of it.

lionelkameni: Hi David,
I was thinking of using enum as I wasnt able to push into piece = Vec::<(u
size, String>)::new (). Compiler kept complaining whenever I tried
piece.push(0,"hello). I got the subtract method on a string using
"hello".to_str().as_bytes.iter ().skip ().take ().collect().
But I am more worried converting back to string is an o (n) operation.  I'd
like to use map but that doesn't seem to work

lionelkameni:
Conflict fixed, small issue with the test pending, please do not merge.

src/lib.rs
```diff
@@ -488,22 +497,46 @@ mod test {
...
+    let size = 3 as u64;
+    let the_string = random_string(size);
+    let mut data_map = datamap::DataMap::None;
+  {  
     let mut se = SelfEncryptor::new(&mut my_storage as &mut Storage, datamap::DataMap::None);
-    let the_string = random_string(3);
     se.write(&the_string, 5u64);
-    assert_eq!(se.file_size, 8u64);
-    let data_map = se.close();
+    assert_eq!(se.file_size, (size + 5) as u64);
```
benjaminbollen: Why is this +5?

lionelkameni: Hi ben,
We have a position offset.
Best wishes

src/lib.rs
```diff
@@ -770,41 +804,46 @@ mod test {
   }
 
   #[test]
-  fn check_large_file_size_over_11_chunks() {
-    // has been tested for 50 chunks
+  fn check_10_and_a_bit_max_chunks() {
```
benjaminbollen: How did it get to overwrite a function?

src/lib.rs
```diff
@@ -770,41 +804,46 @@ mod test {
...
-      }
-      assert_eq!(se.get_chunk_size(number_of_chunks), MIN_CHUNK_SIZE);
-      assert_eq!(se.get_next_chunk_number(number_of_chunks), 0);
-      assert_eq!(se.get_previous_chunk_number(number_of_chunks), number_of_chunks - 1);
-      assert_eq!(se.get_start_end_positions(number_of_chunks).0, 
-                        number_of_chunks as u64 * MAX_CHUNK_SIZE as u64);
-      assert_eq!(se.get_start_end_positions(number_of_chunks).1, 
-                        ((number_of_chunks * MAX_CHUNK_SIZE) as u64 + 1024));
+      assert_eq!(se.get_num_chunks(), 11);
+      assert_eq!(se.get_chunk_size(0), MAX_CHUNK_SIZE);
```
benjaminbollen: Here you explicitly replace a generalised loop with an incomplete, unordered set of assertions.  Not only missing out some iterations, but equally the final chunk which is a special case.

src/lib.rs
```diff
@@ -488,22 +497,46 @@ mod test {
     assert_eq!(data, xor(&xor(&data,&pad), &pad));
   }
 
+
+
   #[test]
-  fn check_write() {
+  fn check_write_and_read() {
```
benjaminbollen: why are you combining a more simple test 'check_write' into a 'check_write_and_read' into a single test?

src/lib.rs
```diff
@@ -150,13 +153,13 @@ impl<'a> SelfEncryptor<'a> {
...
   }
 
-  /// This is an implementation of the get_storage function from example
+  /// This is an implementation of the get_storage function from example.
   pub fn get_storage(&'a mut self) -> &'a mut Storage { self.storage }
 
-  /// Write method mirrors a posix type write mechanism
-  /// loosely mimics filsystem interface for easy connection to FUSE like
+  /// Write method mirrors a posix type write mechanism.
+  /// It losely mimics a filesystem interface for easy connection to FUSE like
```
benjaminbollen: I already corrected this to 'loosely', and as a matter of style, it is okay to have elliptic descriptions in the benefit of faster readbility and less text - as a pure matter of taste maybe

src/lib.rs
```diff
@@ -166,9 +169,9 @@ impl<'a> SelfEncryptor<'a> {
     }
   }
 
-  /// the returned content is read from the specified position with specified length
-  /// trying to read beyond the file size will cause the self_encryptor to be truncated up
-  /// and return content filled with 0u8 in the gapping area
+  /// The returned content is read from the specified position with specified length.
+  /// Trying to read beyond the file size will cause the self_encryptor to be truncated up
+  /// and return content filled with 0u8 in the gaping area.
```
benjaminbollen: It is a 'gapping area' not a 'gaping area'

Cargo.toml
```diff
@@ -14,4 +14,6 @@ homepage = "http://www.maidsafe.net"
 rand = "*"
 rust-crypto ="*"
 tempdir = "*"
+docopt_macros = "*"
+rustc-serialize = "*"
```
benjaminbollen: rustc-serialize is a dependecy of rust-crypto, we don't use it ourself directly - I think

src/lib.rs
```diff
@@ -488,22 +497,46 @@ mod test {
     assert_eq!(data, xor(&xor(&data,&pad), &pad));
   }
 
+
+
   #[test]
-  fn check_write() {
+  fn check_write_and_read() {
     let mut my_storage = MyStorage::new();
+    let size = 3 as u64;
```
benjaminbollen: you define it as an immutable variable, but what is the value of that if the code explicitly depends on this value being 3.  This btw is the answer to the rhetorical question below.

src/lib.rs
lionelkameni: Hi ben just send a commit to Dave to have it back in ok

Cargo.toml
lionelkameni: HI Ben,
I need serialize for usage of the docopts library.
Best Wishes

src/lib.rs
benjaminbollen: My bad, thx

PR #18, opened 2015-03-20
Close now moves the SelfEncryptor 

Ryman:
Once a non-Copy value is moved in Rust, it cannot be used again, this should be a lot less error prone than checking bools in function preludes. There doesn't seem to be any code allowed to do anything after a `close` so I think this is a legit fix, but let me know if I've missed something subtle!


lionelkameni:
This might have been why the disk interface tests had ownership issues. I
will retry with the fix. Thanks Rhyman

PR #21, opened 2015-03-26
Replace strings

benjaminbollen:


src/lib.rs
```diff
@@ -160,23 +161,27 @@ impl<'a> SelfEncryptor<'a> {
   /// It losely mimics a filesystem interface for easy connection to FUSE like
   /// programs as well as fine grained access to system level libraries for developers.
   /// The input data will be written from the specified position (starts from 0).
-  pub fn write(&mut self, data: &str, position: u64) {
+  pub fn write(&mut self, data: &Vec<u8>, position: u64) {
```
Ryman: This should be `&[u8]`

src/lib.rs
```diff
@@ -160,23 +161,27 @@ impl<'a> SelfEncryptor<'a> {
...
   /// The returned content is read from the specified position with specified length.
   /// Trying to read beyond the file size will cause the self_encryptor to be truncated up
   /// and return content filled with 0u8 in the gaping area.
-  pub fn read(&mut self, position: u64, length: u64) -> &str {
+  pub fn read(&mut self, position: u64, length: u64) -> Vec<u8> {
     self.prepare_window(length, position, false);
 
-    let raw = &self.sequencer[position as usize..(position + length) as usize];
-    str::from_utf8(raw).unwrap()
+    let mut read_vec : Vec<u8> = Vec::with_capacity(length as usize);
```
Ryman: Can this not just return `&self.sequencer[position as usize..(position + length) as usize]` and therefore return `&[u8]`? (This would leave it up to the caller to decide if they need to allocate or not)

src/lib.rs
```diff
@@ -444,8 +449,8 @@ impl<'a> SelfEncryptor<'a> {
 mod test {
   use super::*;
 
-  fn random_string(length: u64) -> String {
-        (0..length).map(|_| (0x20u8 + (super::rand::random::<f32>() * 96.0) as u8) as char).collect()
+  fn random_bytes(length: u64) -> Vec<u8> {
```
Ryman: If this is now random bytes, is there any requirement for them to still be in the range 0x20..0x20+96?

tests/lib.rs
```diff
@@ -46,8 +46,8 @@ fn data_map_content_only(){
   assert!(dm.has_chunks() == false);
 }
 
-fn random_string(length: u64) -> String {
-  (0..length).map(|_| (0x20u8 + (rand::random::<f32>() * 96.0) as u8) as char).collect()
+fn random_bytes(length: u64) -> Vec<u8> {
+  (0..length).map(|_| (0x20u8 + (rand::random::<f32>() * 96.0) as u8) as u8).collect()
```
Ryman: Same here, also it looks like most of the code could just take `usize` instead of `u64` as you can't have a `Vec` with > `usize::MAX` elements anyway.

src/lib.rs
benjaminbollen: That is a very good suggestion; and it would be immutable, protecting the sequencer by Rust?

src/lib.rs
benjaminbollen: No you are right, and I already changed that in my next pull request.  thx!

tests/lib.rs
benjaminbollen: good point ! thx again

benjaminbollen: should positions then also better be `usize` everywhere?

benjaminbollen: in random_bytes it for sure already makes sense to use `usize`, but we will have to come back to this after v0.1.0 to use memory mapped files to allow for bigger file sizes... I think :)

src/lib.rs
Ryman: Not sure what you're asking with 'protection' but it would just be a view into the sequencer, as far as rust's guarantees go, it should be illegal to edit the sequencer when this view is active without allocating a copy.

You can test this if you do `let foo = bar.read(..); bar.read(..)`, the borrowchecker *should* complain that foo is still live.

tests/lib.rs
Ryman: I think there needs to be some consideration, I'm assuming the system allows files > 4gb, so there needs to be a mechanism for loading these files on a 32bit machine, using usize/vec will not allow this afaik.

But yeah, positions/lengths should be usize as that's equivalent to `size_t` 'guaranteed to be big enough for the size of the biggest object your system can handle'.

benjaminbollen:
Has been pulled in by TestsCleanUp, so addressing the tests/lib.rc build error there

PR #25, opened 2015-03-27
Full tests - part 2

benjaminbollen:


dirvine:
Sorry @benjaminbollen  Conflicts

tests/lib.rs
```diff
@@ -29,6 +29,8 @@ use std::fs::File;
 use std::io::*;
 use tempdir::TempDir as TempDir;
 use std::string::String as String;
+use std::vec::Vec as Vec;
```
Ryman: Why are you importing `Vec` and `String`?

tests/lib.rs
```diff
@@ -85,48 +87,91 @@ impl Storage for MyStorage {
...
   let mut read_position : usize = 0;
-  let mut index : usize = 0;
+  let mut content_len : usize = 4 * MAX_CHUNK_SIZE as usize;
   let mut my_storage = MyStorage::new();
   let mut data_map = datamap::DataMap::None;
-  let mut original : Vec<u8> = Vec::with_capacity(DATA_SIZE as usize);
-  let mut decrypted : Vec<u8> = Vec::with_capacity(read_size);
-  original = random_bytes(DATA_SIZE as usize);
+  let mut original : Vec<u8> = Vec::with_capacity(content_len);
+  original = random_bytes(content_len);
```
Ryman: This can just be `let original = random..`, there's an allocation already within random_bytes.

tests/lib.rs
```diff
@@ -85,48 +87,91 @@ impl Storage for MyStorage {
...
-    // read beyond the file, output is padded with default initalisation
-    read_position = DATA_SIZE as usize - read_size + 2000;
-    let decrypted = se.read(read_position as u64, read_size as u64);
-    let mut padded : Vec<u8> = Vec::with_capacity(read_size);
-    padded.push_all(&original[read_position..DATA_SIZE as usize]);
-    padded.resize(read_size, 0u8);
-    assert_eq!(padded, decrypted);
+    {
+      let mut decrypted : Vec<u8> = Vec::with_capacity(read_size);
+      let decrypted = se.read(read_position as u64, read_size as u64);
```
Ryman: Same for here, there's no need for initial `decrypted` vec.

tests/lib.rs
benjaminbollen: This was there, not sure who imported it, I had it removed.

tests/lib.rs
benjaminbollen: Is it not desirable to have the capacity of the vector set explicitly?

tests/lib.rs
benjaminbollen: noted

tests/lib.rs
Ryman: But you're returning a `Vec` so this isn't using the allocation from the first line, it's creating a new one and assigning to the slot.

dirvine: Yes @benjaminbollen this is accurate, thanks @Ryman really appreciate this help, it's fantastic and helping the guys a lot. Owe ye a beer :-)

PR #29, opened 2015-03-30
Full tests - part 3

benjaminbollen:
Mostly clean up of warnings,

some dead code remains on temp dir and Chunk location `OnDisk`


dirvine:
Can you please remove all dead code now Ben, it was for a later iteration but won't be needed with MemorMapped type for vector. So lets go 100% warning free. A nice success

benjaminbollen:
OK! yes, I wanted to ask you that question this morning, but missed the opportunity

PR #31, opened 2015-03-31
WIP Example

benjaminbollen:


tests/lib.rs
```diff
@@ -173,13 +175,14 @@ fn write_random_sized_out_of_sequence_writes_with_gaps_and_overlaps() {
...
+      let offset : usize = offsets[ind];
+      total_size = std::cmp::max(total_size, offset+ piece_size);
       se.write(pieces[ind], offset as u64);
-
-
+      assert!(DATA_SIZE >= total_size as u64);
+      let decrypted = se.read(offset as u64, piece_size as u64);
+      assert_eq!(decrypted, original[offset..(offset + piece_size)].to_vec());
     }
+    //assert_eq!(total_size as u64, se.len());
```
dirvine: please uncomment or remove.

benjaminbollen: This slipped into this PR, but I'd need to clean it up.  The example PR is not indented to touch tests/lib.rs.  Will merge latest master

Cargo.toml
```diff
@@ -14,6 +14,13 @@ homepage = "http://www.maidsafe.net"
...
 rust-crypto ="*"
 tempdir = "*"
-docopt_macros = "*"
-rustc-serialize = "*"
 
+[lib]
+name = "self_encryption"
+bench = false
+
+[[bin]]
```
Ryman: This is unnecessary, just put se.rs in /examples and then you can use `cargo run --example se -- args` (might be a little off since it's from memory but the gist of it should be correct)

examples/self_encryption/main.rs
```diff
@@ -0,0 +1,138 @@
...
+  source_path : &'b path::Path,
+  name : &'b str;
+  dir : &'b path::Path,
+  destination_path : &'b path::Path
+}
+
+
+impl MyStorage {
+  pub fn new(filepath_ : &str) -> MyStorage {
+    MyStorage { source_path : match path::Path::new(filepath_) {
```
Ryman: Formatting here is really hard to read, better to extract temporaries

Cargo.toml
benjaminbollen: oh okay, I took this from another crate. cheers

examples/self_encryption/main.rs
```diff
@@ -0,0 +1,138 @@
...
+    let mut s = String::new();
+    //f.read_to_string(&mut s); put f into a string
+    match f.read_to_string(&mut s){
+        Err(why) => panic!("on get couldn't read: {}", why),
+        Ok(_) => print!("contains:\n{}", s),
+    }
+    s.into_bytes()
+  }
+
+  #[allow(unused_must_use)]
```
Ryman: The form `let _ = some_must_use();` is preferred as a reader can know where an ignored value is.

examples/self_encryption/main.rs
benjaminbollen: I'm really bleeding here... I trying to struggle through but I do not know what I'm doing with Rust and file system :)!

benjaminbollen:
Ran into grid lock, took a step back an progressing in PR Simple Example (which will do the same thing)

PR #40, opened 2015-04-10
Some fixes

steveklabnik:
First commit: extra extern crates that weren't needed

Second commit: basically stylistic fixes

### DIFF
--- a/src/datamap.rs
+++ b/src/datamap.rs
@@ -16,8 +16,6 @@
 // See the Licences for the specific language governing permissions and limitations relating to
 // use of the MaidSafe Software.
 
-extern crate rustc_serialize;
-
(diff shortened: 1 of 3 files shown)


dirvine:
Welcome and huge thanks for this PR 
This is causing a failure in test ( https://travis-ci.org/dirvine/self_encryption/jobs/57998935#L1268 ) due to a rand call there (only in the test though).

steveklabnik:
Oh! Yeah, I didn't have the prereqs, so I didn't _run_ the tests, but I should be able to fix it...

steveklabnik:
Check that new commit. That's generally the Right Way (tm)

dirvine:
Excellent Steve Thanks for that.

PR #56, opened 2015-05-01
Compile examples as part of cargo test

rossmuir:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/self_encryption/56)


Ryman:
I think this is already the default behavior?

If you add random text before the imports in the example and then run `cargo test` you already get a compile failure? What's the benefit of this?

dirvine:
Thx
I think it was marked false by one of the devs in an earlier commit. It is default to have these build the tests, so yes arguably leaving the line blank is the same effect.

PR #57, opened 2015-05-01
Add example instructions

rossmuir:
[Review on Reviewable](https://reviewable.io/reviews/dirvine/self_encryption/57)

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ A version of [convergent encryption](http://en.wikipedia.org/wiki/Convergent_enc
 
 #Basic self_encryptor example DIY instructions
 
-Instructions to use the 'basic_encryptor' example to encrypt a file into encrypted chunks and a datamap and then decrypt the datamap and restore the file.
+Instructions to use the 'basic_encryptor' example. This library splits a file into encrypted chunks producing a data map. The correct credentials decrypt the data map enabling the file to be reconstituted. 
 
 1. install rust Nightly (the compiler) http://www.rust-lang.org/install.html or on mac / linux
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -46,3 +46,28 @@ A version of [convergent encryption](http://en.wikipedia.org/wiki/Convergent_enc
 
 #Video of the process 
 [self_encryption process and use case video] (https://www.youtube.com/watch?v=Jnvwv4z17b4)
+
+#Basic self_encryptor example DIY instructions
+
+Instructions to use the 'basic_encryptor' example to encrypted a file into encrypted chunks and a datamap and then decrypted the datamap and restore the file.
```
Viv-Rajkumar: `to encrypted a file` should prolly be `to encrypt a file` similar for `decrypted` -> `decrypt`. Maybe give @NickLambert  a shout :)

rossmuir: Cheers @Viv-Rajkumar  updated

NickLambert: Suggest change from : Instructions to use the 'basic_encryptor' example to encrypted a file into encrypted chunks and a datamap and then decrypted the datamap and restore the file.

to 

Instructions to use the 'basic_encryptor' example. This library splits a file into encrypted chunks producing a data map, which is also split into chunks and encrypted.  The correct credentials decrypt the data map enabling the file to be reconstituted.

rossmuir: Cheers Nick I updated the PR with your feedback but omitted the "which is also split into chunks and encrypted" part as I don't believe the example does this part. Thanks :-)

dirvine:
Nice one Ross, thx very much. will help community a lot.

PR #59, opened 2015-05-07
Parallel encryption

bcndanos:
Two days ago I had created a new module in crates.io called "asynchronous". It allows parallel and deferred execution of threads in a easy and safe way.
By now, I modified the encryption part of the project to adapt to this module. Now the encryption process runs in all the CPU cores.

[Review on Reviewable](https://reviewable.io/reviews/dirvine/self_encryption/59)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -17,6 +17,7 @@ docopt = "*"
 rustc-serialize = "*"
 rust-crypto = "*"
 cbor = "*"
+asynchronous = "*"
 
 [[example]]
 name = "basic_encryptor"
--- a/examples/basic_encryptor.rs
+++ b/examples/basic_encryptor.rs
@@ -30,6 +30,7 @@ use std::io::prelude::*;
 use std::path::Path;
 use std::string::String;
 use std::error::Error;
+use std::sync::Arc;
 
 use docopt::Docopt;
 use cbor::{ Encoder, Decoder};
@@ -97,7 +98,7 @@ impl Storage for MyStorage {
     data
   }
 
-    fn put(&mut self, name: Vec<u8>, data: Vec<u8>) {
+    fn put(&self, name: Vec<u8>, data: Vec<u8>) {
     let pathstr = file_name(&name);
     let tmpname = self.storage_path.clone() + &pathstr;
     let path = Path::new(&tmpname);
@@ -123,7 +124,7 @@ fn main() {
         Err(why) => println!("! chunk_store_test {:?}", why.kind()),
         Ok(_) => {},
     }
-    let mut my_storage = MyStorage { storage_path : "chunk_store_test/".to_string() };
+    let my_storage = Arc::new(MyStorage { storage_path : "chunk_store_test/".to_string() });
     
     if args.flag_encrypt && args.arg_target.is_some() {
         let mut file = match File::open(&args.arg_target.clone().unwrap()) {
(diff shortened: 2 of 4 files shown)


dirvine:
Well, thank you very much, it will be reviewed very soon. Welcome to the project, really nice.

benjaminbollen:
Ran a benchmark on my laptop; of this PR and maidsafe/async branch both merged into master: there is an overall few MB/s increase in speed; but more importantly: very notably the stress is no longer on a single core, but spread over the cores.  So very nice work.  I have not dived into the code yet.

with async: 
running 5 tests
test bench_write_then_read_a_200B ... bench:     13067 ns/iter (+/- 938) = 30 MB/s
test bench_write_then_read_b_1KB  ... bench:     47447 ns/iter (+/- 1313) = 43 MB/s
test bench_write_then_read_c_1MB  ... bench:  53708306 ns/iter (+/- 1767000) = 37 MB/s
test bench_write_then_read_d_3MB  ... bench: 157556196 ns/iter (+/- 3227794) = 37 MB/s
test bench_write_then_read_e_10MB ... bench: 523736052 ns/iter (+/- 21033036) = 20 MB/s

master (without async):
running 5 tests
test bench_write_then_read_a_200B ... bench:     13838 ns/iter (+/- 982) = 28 MB/s
test bench_write_then_read_b_1KB  ... bench:     49494 ns/iter (+/- 1936) = 41 MB/s
test bench_write_then_read_c_1MB  ... bench:  56313256 ns/iter (+/- 843919) = 35 MB/s
test bench_write_then_read_d_3MB  ... bench: 170048836 ns/iter (+/- 4660920) = 31 MB/s
test bench_write_then_read_e_10MB ... bench: 588400238 ns/iter (+/- 48747987) = 20 MB/s

bcndanos:
I suppose you have modified the bench.rs for include the Storage struct as "Arc" and "Mutex". I forgot it in the PR.
In that case, the part where the tests "put" to storage don't run in parallel, the threads have a mutual exclusión.
The implementation of Storage Trait in the example/basic_encryptor.rs is even better, aprox. 40%, due to write to disk in parallel.

benjaminbollen:
Nope I did not ;-) let me give that a run for its money :)

benjaminbollen:
The basic_encryptor example indeed gives a notable speed up on the disk IO.  for a 185MB file with the async write out, it shaves off almost 1 minute down to 2min28 on my machine.  So very nice indeed!  What we need next is the implementation to asynchronously SE, because that process still happens synchronously.

bcndanos:
Should I do anything else?

BSMaidSafe:
Hi, Having updated the bench tests as you suggested there's a noticeable performance improvement there also. I'll put in a PR for that, and recommend to David that both can be merged. Thanks again for your contribution. Out of curiosity, do you have any plans for extending, or otherwise maintaining the asynchronous crate?

bcndanos:
Yes, I'm improving the asynchronous crate continuosly. 
Just now I updated to versión 0.4.1 that allows EventLoops. I'm testing this change with Crust module, in beacon.rs and works perfectly seeking peers. In this case, it doesn't bring any benefit in terms of computational time, but reduces the code significantly.
If you need any change, just PR any change to me.

BSMaidSafe:
Excellent, good to know. Will do on the PR's if required.

dirvine:
Very nice work, thank you very much. I See this being very useful in crust, not to help beacon so much but for parallel attempts (up to 15 at a time) to bootstrap from the bootstrap file. This is sorely missing yet and on the roadmap, if you fancy trying that one. Then @chandraprakash will buy you a huge beer.

bcndanos:
I've been reviewing at the code and I think it's easy to do it. But before I have to develop a new method in my module for processing in parallel until at least one task(in this case one connection) is done.

PR #70, opened 2015-05-19
Adds de/compression to SelfEncryptor.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/70)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,7 @@ rustc-serialize = "*"
 rust-crypto = "*"
 cbor = "*"
 asynchronous = "*"
+flate2 = "0.2"
 
 [[example]]
 name = "basic_encryptor"
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -100,13 +100,22 @@ extern crate rand;
 extern crate crypto;
 extern crate rustc_serialize;
 extern crate asynchronous;
+extern crate flate2;
+
 use std::sync::Arc;
 use std::cmp;
 use crypto::sha2::Sha512 as Sha512;
 use crypto::digest::Digest;
+use crypto::symmetriccipher::SymmetricCipherError::{InvalidLength, InvalidPadding};
 use std::iter::repeat;
 use asynchronous::Deferred;
 use asynchronous::ControlFlow;
+use std::io::Write;
+use std::io::Read;
+use std::error::Error;
+use flate2::write::GzEncoder;
+use flate2::read::GzDecoder;
+use flate2::Compression;
 
 // This is pub to test the tests directory integration tests; these are temporary and need to be
 // replaced with actual integration tests. This should be private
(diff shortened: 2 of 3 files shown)


maidsafe-highfive:
r? @Fraser999

dirvine:
r? dirvine

dirvine:
Thx Brian can you confirm the example still operates effectively and what impact this makes on benchmarks

s-brian:
Yes, the example still works. There's no noticeable impact on the integral values returned from benchmarks.

dirvine:
Yes seen same locally, nice one man, cheers for this. I think we may still need a test to check it is deterministic across OS's. So a chunk printout and confirm we get that chunk from a known input stream. If the test passes across OS's then we are sorted. IT will confirm AES is deterministic to which is really important for de-duplication and cross os files. 

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.
Reviewed files:
- Cargo.toml @ r1
- src/lib.rs @ r1
- tests/lib.rs @ r1

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

s-brian:
Are we looking to create a file whose content is encrypted and whose pre/hashes are written to another file. The test would then involve encrypting the file content and checking the hashes against those written to the other file. Or maybe a large string and pre/hashes located in the test file to be encrypted and checked in a test?

dirvine:
Yes a large string (can be a repeated abc type thing, with known chunk outputs is fine. Just to confirm all OS's produce deterministic and same output from the internal algorithms. 

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

dirvine:
Should be an Integration tests (in the tests/ dir I think)

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

s-brian:
Ok, cheers.

dirvine:
No prob, sorry to be a PITA, but we can also tick of 2 TODO items with this commit as well. Great progress.

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

dirvine:
I will update crates versions as well when you are done. At least I will package and publich if you update the toml

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

s-brian:
No worries. Is it a patch increase?

dirvine:
Yes I think so Brian, cheers. 

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

s-brian:
Cool, thanks.

dirvine:
r? @dirvine

---

Review status: all files reviewed, all discussions resolved, some commit checks failed.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/70)

dirvine:
Was there an issue with determinism Brian or another reason for close?

s-brian:
It's got to do with Unicode again, although the test appears to be ok, and I've written an alternative with ANSI characters that's ok. If I create a 'String' using a valid range of Unicode characters and try to print it the test panics saying it's not a valid Unicode string! I'd like to get the bottom of that before doing a PR, unless your happy to accept the PR and we can look into it later. As it stands the test passes on Windows and generates 3 unique hashes from the input strings that can be tested on other OS's.

dirvine:
I would say if compression does not cause issues but we have a potential unicode handling issue then this PR is OK and we can add a task to confirm unicode. So unless compression has caused this issue I would merge this PR (if it passes on all OS's) and fix unicode next. IF that makes sense then lets go that way.

s-brian:
I'll do a PR and take it from there. Cheers David.

PR #72, opened 2015-05-20
Compression and deterministic test.

s-brian:


s-brian:
That settles it, I'll close the PR and take a look on OSX/Linux.

dirvine:
Cool :+1:

dirvine:
worth testing without compression to check it#s not just that.

s-brian:
With compression only it was fine, so looks like the deterministic test is not deterministic at the moment.

Issue #81, opened 2015-06-11
Revert rust-crypto


Fraser999:
We need to revert rust-crypto to version "0.2.30" since it has a critical issue in Release mode.

Fraser999:
Turns out the fix wasn't correct, so this will need to be handled differently.

Rather than reverting rust-crypto, the current suggestion is to try and replace its use with sodiumoxide.

PR #82, opened 2015-06-11
Replaced use of rust-crypto with sodiumoxide.  Applied "ignore" attri…

Fraser999:
…bute to a test which is frequently exposing at least one bug in SelfEncrypt.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/82)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,7 +7,6 @@ build-tests/
 target/
 *~
 /.idea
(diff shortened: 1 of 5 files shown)


Fraser999:
Fixes #81  Exposed #83 hence the test is marked as "ignore" for now.

PR #98, opened 2015-07-29
using travis osx and scripts

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/98)


maidsafe-highfive:
r? @s-brian

PR #101, opened 2015-08-03
Deflate compression/decompression.

s-brian:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/101)


maidsafe-highfive:
r? @maqi

dirvine:
Looks good but fails on beta/stable, can we confirm why and see if we can improve on that. Otherwise the compression is a success by looks of it.

s-brian:
The benchmarks use unstable #![feature(test)].

PR #125, opened 2015-10-31
Corrected bad last chunk size (in get_chunk_size) and added a test fu…

Thierry61:
I wanted to self-encrypt a vector serialized with cbor and I came across an error when I tried to deserialise it.

The problem was reproduced in the test function I have added (check_serialised_vector). The generated error was:

>thread 'test::check_serialised_vector' panicked at 'failed to deserialise Vec<usize>'

I investigated and found that the sum of source_size field of the chunks was differing from the file size (one byte difference). And then I found the error in the following formula computing the size of last chunk (< 3 * MAX_CHUNK_SIZE case in get_chunk_size function):

(2 \* **(**self.file_size / 3**)**)

The highlighted parenthesis’s were missing, meaning that multiplication by 2 was done before division by 3, which is wrong because we want to subtract the size of the first 2 chunks.

With this correction the new test function is ok.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/125)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -727,7 +727,7 @@ impl<S:Storage + Send + Sync + 'static> SelfEncryptor<S> {
             if chunk_number < 2 {
                 return (self.file_size / 3) as u32
             } else {
-                return (self.file_size - (2 * self.file_size / 3)) as u32
+                return (self.file_size - (2 * (self.file_size / 3))) as u32
             }
         }
         if chunk_number < self.get_num_chunks() - 2 {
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @brian-js (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. The way GitHub handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Please see [CONTRIBUTOR.md](../blob/master/CONTRIBUTOR) for more information.

dirvine:
Yes, very nice, this should be counted as a 4 point bounty (2 for bug 2 for tests to maintain the code) . I will ping @NickLambert  Can you please supply a btc address (if you wish). Many thanks

Thierry61:
Many thanks for the offered bounties. I accept them with great pleasure. Here is my BTC address:
16AXRezDRtg454pGEQUjozZQx4YC1EFw8D

NickLambert:
Payment of 0.2555 BTC on the way. Thanks for your help @Thierry61!

Thierry61:
Payment has been confirmed. Thanks for the bounty.

I have added a new commit that improves my test function by serializing vectors of various sizes.

I have also modified all existing tests functions by checking that after a write the file size has the expected value, in both:

- `file_size` field of the self-encryptor itself
- cumulated value of `source_size` field of all chunks

PR #137, opened 2015-11-18
switch to use memmap from crates

maqi:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/137)


maidsafe-highfive:
r? @brian-js

PR #145, opened 2016-01-11
Simplify get_start_end_positions

lgvz:
Also, operands are cast to u64 before multiplication to avoid problems
with large files.

(I removed the reviewable.io banner because I don't like it when crap is injected into *my* comment as if *I* was somehow promoting something.)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -763,17 +763,12 @@ impl<S:Storage + Send + Sync + 'static> SelfEncryptor<S> {
             return (0, 0)
         }
         let start;
-        let penultimate = (self.get_num_chunks() - 2) == chunk_number;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
Thanks for the pull request, and welcome! The MaidSafe team is excited to review your changes, and you should hear from @maqi (or someone else) soon.

If any changes to this PR are deemed necessary, please add them as extra commits. This ensures that the reviewer can see what has changed since they last reviewed the code. The way GitHub handles out-of-date commits, this should also make it reasonably obvious what issues have or haven't been addressed. Large or tricky changes may require several passes of review and changes.

Please see [CONTRIBUTOR.md](../blob/master/CONTRIBUTOR) for more information.

PR #151, opened 2016-01-15
Test multiple subsequent write calls

lgvz:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/151)


maidsafe-highfive:
r? @brian-js

PR #155, opened 2016-01-20
Nits

lgvz:
Fixed just some little annoying things...

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/155)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -196,11 +196,11 @@ impl Sequencer {
     }
 
     /// Initialise as a memory map
-    pub fn as_mmap() -> Sequencer {
-        Sequencer {
+    pub fn as_mmap() -> Result<Sequencer> {
+        Ok(Sequencer {
             vector: None,
-            mmap: Some(Mmap::anonymous(MAX_MEMORY_MAP_SIZE, Protection::ReadWrite).unwrap()),
-        }
+            mmap: Some(try!(Mmap::anonymous(MAX_MEMORY_MAP_SIZE, Protection::ReadWrite))),
+        })
     }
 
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

src/lib.rs
```diff
@@ -389,13 +389,13 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
         if file_size <= MAX_IN_MEMORY_SIZE as u64 {
             sequencer = Sequencer::as_vector();
         } else {
-            sequencer = Sequencer::as_mmap();
+            sequencer = Sequencer::as_mmap().unwrap();
```
dirvine: Another nice PR, this unwrap though is probably not good to leave. I think this may mean pushing Results up, which is OK and better than unwrap.

src/lib.rs
```diff
@@ -1022,7 +1021,7 @@ mod test {
         // check read, write
         let mut new_se = SelfEncryptor::new(my_storage.clone(), data_map);
         let fetched = new_se.read(0, MIN_CHUNK_SIZE as u64 * 3);
-        assert_eq!(fetched, the_bytes);
```
dirvine: Not sure why removing the eq! is done here?

lgvz: Because when assert_eq! fails it will print the left and right args so in this case the contents of the vectors - that's quite a screenful.

src/lib.rs
lgvz: Yes, my change here is just a first step towards proper error management. Maybe in another PR...

PR #158, opened 2016-01-22
Added clippy.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/158)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -109,6 +109,11 @@
 #![allow(box_pointers, fat_ptr_transmutes, missing_copy_implementations,
          missing_debug_implementations, variant_size_differences)]
 
+#![cfg_attr(feature="clippy", feature(plugin))]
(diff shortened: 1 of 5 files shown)


maidsafe-highfive:
r? @brian-js

src/lib.rs
```diff
@@ -446,7 +445,7 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
             }
         }
         self.prepare_window(data.len() as u64, position);
-        for i in 0..data.len() {
+        for (i, _) in data.iter().enumerate() {
             self.sequencer[position as usize + i] = data[i];
```
lgvz: How about:

    for (i, &byte) in data.iter().enumerate() {
        self.sequencer[position as usize + i] = byte;

Fraser999: Good point! :)

PR #159, opened 2016-01-22
Fix truncate

lgvz:
Truncation didn't work for similar reasons as writing.

Another problem was that `prepare_window` didn't make sure sequencer had enough room for the content of the last chunk to be decrypted. And sure enough, this resulted in index-out-of-bounds when initializing `SelfEncryptor` with a `DataMap::Chunks` - which leaves sequencer's length to zero - and then truncating.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/159)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -564,26 +564,33 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
         }
     }
 
-    /// Truncate the self_encryptor to the specified size (if extend, filled with 0u8).
-    pub fn truncate(&mut self, position: u64) -> bool {
+    /// Truncate the self_encryptor to the specified size (if extended, filled with 0u8).
+    //
+    // NOTE: Right now calls prepare_window with the full length of the truncated file
+    // because chunk sizes may change. Chunks that don't change don't actually need any
+    // treatment here so there's room for performance improvement, especially for large
+    // files.
+    pub fn truncate(&mut self, new_size: u64) -> bool {
+        if self.file_size == new_size {
+            return true;
+        }
         let old_size = self.file_size;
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

src/lib.rs
```diff
@@ -564,26 +564,33 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
...
-        if position < old_size {
-            self.sequencer.truncate(position as usize);
-            let last_chunk = self.get_chunk_number(position) + 1;
-            self.chunks.truncate(last_chunk as usize);
+        if new_size < old_size {
+            self.prepare_window(new_size, 0);   // call before changing self.file_size
+            self.file_size = new_size;
+            self.sequencer.truncate(new_size as usize);
+            let num_chunks = self.get_num_chunks(); // call with updated self.file_size
+            self.chunks.truncate(num_chunks as usize);
```
lgvz: This is flaky because `self.chunks` is not guaranteed to be sorted. This lib has a lot of problems...

Fraser999: @lgvz Agreed.  Would it be possible for you to contact me directly to discuss this further?  fraser.hutchison@maidsafe.net

lgvz:
It's me: 34gv...ogw1

dirvine:
:+1:  All the info we need for this stuff, so cool and thx again. @NickLambert will sort out some btc as a thanks from us and the community you are helping here.

lgvz:
Thank you! :)

NickLambert:
Thanks @lgvz, 0.4 BTC winging its way to you.

PR #160, opened 2016-01-24
Avoid re-encrypting chunks if file has not been modified.

Fraser999:
[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/160)


maidsafe-highfive:
r? @brian-js

lgvz:
`fn close` will have to be changed also, see #163.

lgvz:
This will fail the test added by #163...

brian-js:
@lgvz Shall I fix it in a separate PR, or would you be ok do it in #163?

lgvz:
@brian-js whatever works for you.

brian-js:
Ok, I'll push a fix that'll allow your new test to pass.

PR #162, opened 2016-01-24
Make Storage::get take name as a slice instead of Vec

lgvz:
Caller can avoid needless clones.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/162)


maidsafe-highfive:
r? @brian-js

PR #163, opened 2016-01-27
Add test for overwriting when started with existing datamap

lgvz:
This is designed for PR #160. If a chunk and therefore its hash changes, then also the next two chunks need to be re-encrypted.

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/163)


maidsafe-highfive:
r? @brian-js

PR #166, opened 2016-02-01
Some rework on SelfEncryptor

lgvz:
- Only the chunks of the original datamap are tracked until close is
  called.
- Datamap is sorted just once, in SelfEncryptor::new.
- When a whole range of chunks are overwritten, the middle chunks are
  not decrypted. And other similar optimizations.
- Some methods are now free functions because I think they are more
  flexible and less error prone. No need to change `self.foo` just to be able
  to call `self.bar` correctly and no implicitly passed args.

The result is simpler and more efficient.

Btw, I changed `fn close(&mut self)` to `fn close(mut self)`. I saw from git log that it used to be this way but was changed to `&mut` in commit c09a7cf but I couldn't find a reason. Did I miss something?

[Review on Reviewable](https://reviewable.io/reviews/maidsafe/self_encryption/166)

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -129,7 +129,6 @@ mod encryption;
 mod datamap;
 
 use std::cmp;
-use std::collections::BTreeSet;
 use std::error::Error;
 use std::fmt::{self, Debug, Formatter};
 use std::io::{self, ErrorKind, Read, Result, Write};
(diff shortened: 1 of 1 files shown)


maidsafe-highfive:
r? @maqi

dirvine:
r? @Fraser999 

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/166#-:-K9lKPqgWqv9lPNVUIQA:-596912412)*

brian-js:
Reviewed 1 of 1 files at r1.
Review status: all files reviewed at latest revision, all discussions resolved.

---



*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/maidsafe/self_encryption/166)*

brian-js:
Thanks @lgvz.

src/lib.rs
```diff
@@ -390,33 +390,39 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
...
 
+        let sorted_map;
+        let chunks;
+        let map_size;
         match datamap {
             DataMap::Content(ref content) => {
                 sequencer.init(content);
-                chunks.push(Chunk {
-                    number: 0,
+                sorted_map = vec![];
```
dirvine: Either name or container type may be misleading here?

src/lib.rs
```diff
@@ -431,176 +437,160 @@ impl<S: Storage + Send + Sync + 'static> SelfEncryptor<S> {
...
-    /// read beyond the file size will cause the self_encryptor to be truncated up and return
-    /// content filled with 0u8 in the gap.  Any other unwritten gaps will also be filled with
+    /// read beyond the file size will cause the self_encryptor to return content filled with 0u8
+    /// in the gap (file size isn't affected).  Any other unwritten gaps will also be filled with
     /// '0u8's.
     #[cfg_attr(feature="clippy", allow(cast_possible_truncation))]
     pub fn read(&mut self, position: u64, length: u64) -> Vec<u8> {
-        self.prepare_window(length, position, false);
-        let mut read = Vec::with_capacity(length as usize);
+        self.prepare_window_for_reading(position, length);
```
dirvine: This is much better

NickLambert:
Thanks for the PR @lgvz. Can you advise which bitcoin address you would like me to send a bounty to?

lgvz:
Oh,  3Ka3tSMQdNABvmSV2fCAnXXgDJdB4UffCM. I appreciate!

NickLambert:
Thanks @lgvz . Just moving some BTC from our vault to our wallet. I'll send some across once this transaction is complete.

NickLambert:
Payment on the way @lgvz . Thanks again!

PR #205, opened 2017-03-24
chore/ci: update rustc and clippy versions

nbaksalyar:


ustulation:
good point - maybe explicitly disable clippy with attirbs for these ones nikita (just don't put a global disable - see case by case).

Or better create a trivial struct available only to test-code such as:
```
#[derive(PartialEq, Eq)]
pub struct Arr<'a>(pub &'a [u8]);
impl<'a> Debug for Arr<'a> {
    fn fmt(f: &Formatter) { write!("{}..{}", self.0[..4], self.0[self.0.len() - 4..]) }
}```

(that is just a hint ^^^ needs to be done properly if doing it this way)
then ensure the code does `assert_eq!(Arr(&vec0), Arr(&vec1))`

Is that better than turning off the lints in many places/using a global allow ? @Fraser999 @nbaksalyar

**Comment by nbaksalyar** [comment]
@ustulation 
>Is that better than turning off the lints in many places/using a global allow ? 

Yes, this definitely sounds like a good tradeoff to me. But I think it can be even simpler, as `assert!` at least allows to use a custom format for output in case of a failure.

**Comment by ustulation** [comment]
@nbaksalyar that's allowed by `assert_eq!` as well - i dont' think it's relevant here though ? It's more about using the lint suppressing attribute.

**Comment by Fraser999** [comment]
I'd go for disabling the warning personally.  From back when I had to debug these test failures, I remember the actual binary data right at the start or end often wasn't the problem - more a missing part in the middle, or a part in the middle which hadn't been overwritten.  I don't feel strongly here though - my main point was to not print the full binary data, so I'd be OK with anything except that :)

**Comment by nbaksalyar** [comment]
@ustulation well, what I mean is something like this:
```
assert_eq!(a, b, "{} does not equal to {}", a[..4], a[a.len() - 4..])
```
That should work basically like your `Arr` sample.

ustulation:
@nbaksalyar not sure i am getting you - that has nothing to do with the assert not printing the entire arrays on panic. That formatter is just the additional stuff it will print. It will still print the entire 2 arrays saying which was the `left` and which was the `right`. Or am i understanding something wrong ?

nbaksalyar:
@ustulation, hmm yeah, you're right, just checked it - for some reason I thought that it'd _replace_ the failure output rather than add more info to it.

Then the custom formatter solution should be OK, I guess? Or do you think it'd be easier to turn off clippy lints for these cases?

ustulation:
i would say go for the custom formatter - a global allow will affect cases it should not and putting the attrib everwhere is more verbose (and worse) if it's in many places. Custom struct means that it's a concious decision and easily to understand why it's done that way if someone looks at its use.

PR #231, opened 2018-05-12
Replace the C wrapper for the brotli compression library with the pure Rust impl [MAID-2651]

nbaksalyar:
This PR replaces `brotli2`, which is a Rust wrapper around the native C library, with `rust-brotli`, a pure-Rust implementation of the Brotli compression algorithm. This change makes `self_encryption` independent of native libs and thus now it can be compiled on any platform supported by the Rust compiler.


maidsafe-highfive:
r? @maqi

PR #234, opened 2018-07-19
Switch from rust_sodium to safe_crypto [MAID-2721]

nbaksalyar:


src/self_encryptor.rs
```diff
@@ -12,9 +12,11 @@ use super::{
 use brotli;
 use brotli::enc::BrotliEncoderParams;
 use data_map::{ChunkDetails, DataMap};
-use encryption::{self, Iv, Key, IV_SIZE, KEY_SIZE};
 use futures::{future, Future};
-use rust_sodium;
+use safe_crypto::{
+    self, hash, Nonce as Iv, SymmetricKey as Key, NONCE_SIZE as IV_SIZE,
```
m-cat: I'm wondering about our function-import rule here. Isn't it out policy not to import `hash` and to use `safe_crypto::hash` instead of just `hash`?

nbaksalyar:
Closed PR - will be superseded by a new one.

PR #238, opened 2018-10-03
[MAID-2721] refactor/all: switch from rust_sodium to safe_crypto 

m-cat:


maidsafe-highfive:
r? @maqi

lionel1704:
This crate has migrated away from rust_sodium in #255

PR #239, opened 2018-12-27
Upgrade to rust 2018 edition

pierrechevalier83:
* Upgrade compiler
* Switch to 2018 edition
* Fix compilation, fmt and clippy warnings
* Remove nightly from CI

### DIFF
--- a/src/sequential/large_encryptor.rs
+++ b/src/sequential/large_encryptor.rs
@@ -9,12 +9,12 @@
 use super::medium_encryptor::MediumEncryptor;
 use super::small_encryptor::SmallEncryptor;
 use super::{utils, SelfEncryptionError, Storage, MAX_CHUNK_SIZE, MIN_CHUNK_SIZE};
-use data_map::{ChunkDetails, DataMap};
+use crate::data_map::{ChunkDetails, DataMap};
+use crate::util::{BoxFuture, FutureExt};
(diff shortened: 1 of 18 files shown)


maidsafe-highfive:
r? @maqi

src/sequential/large_encryptor.rs
```diff
@@ -39,6 +39,7 @@ where
     // Constructor for use with pre-existing `DataMap::Chunks` where there are more than three
     // chunks.  Retrieves the first two and and last two chunks from storage and decrypts them to
     // its internal buffers.
+    #[allow(clippy::new_ret_no_self)]
```
m-cat: Maybe we can allow this crate-wide instead of five separate times throughout the crate. I don't personally like this lint anyway as it fails on a lot of real-world examples -- even the std lib has `new` functions that don't return `Self`.

pierrechevalier83: I was thinking that too. My reasoning was that we could want to encourage future `new` functions to return `Self`, but I'm with you that `new` is a good name for a function that returns a `Result<Self, _>`...
I'll deny it crate-wide.

pierrechevalier83: As discussed, we'll tackle this with the next CI PR (musl)

PR #255, opened 2019-11-27
Remove rust_sodium and replace encryptions with AES

### LINKED ISSUE #254: Deprecation of Rust-Sodium
Replace rust_sodium dependency in self_encryption with [crate aes](https://docs.rs/aes/0.3.2/aes/).
- [AES128](https://docs.rs/aes/0.3.2/aes/struct.Aes128.html) is preferred to be used to replace the encryption mechanisms in the modules wherever rust_sodium is currently used.
- Removal of rust_sodium makes self_encryption a Rust-only crate by moving away from C dependencies.

Yoga07:
Resolves #254


src/encryption.rs
```diff
@@ -8,18 +8,46 @@
 
 // TODO(dirvine) Look at aessafe 256X8 cbc it should be very much faster  :01/03/2015
```
dirvine: We can remove this comment now.

src/encryption.rs
```diff
@@ -8,18 +8,46 @@
...
+// Buffer size is set to max chunk size + 100 bytes for padding.
+pub const BUFFER_SIZE: usize = MAX_CHUNK_SIZE as usize + 100;
 
-pub fn encrypt(data: &[u8], key: &Key, iv: &Iv) -> Vec<u8> {
-    secretbox::seal(data, iv, key)
+pub fn encrypt<E>(data: &[u8], key: &Key, iv: &Iv) -> Result<Vec<u8>, SelfEncryptionError<E>>
+where
+    E: StorageError,
+{
+    let cipher = Aes128Cbc::new_var(key.0.as_ref(), iv.0.as_ref()).unwrap();
```
dirvine: We should now unwrap here `?` is better

src/encryption.rs
```diff
@@ -8,18 +8,46 @@
...
+type Aes128Cbc = Cbc<Aes128, Pkcs7>;
 
-pub use rust_sodium::crypto::secretbox::{Key, Nonce as Iv};
-pub type DecryptionError = ();
+pub const KEY_SIZE: usize = 16;
+pub const IV_SIZE: usize = 16;
 
-pub const KEY_SIZE: usize = KEYBYTES;
-pub const IV_SIZE: usize = NONCEBYTES;
+// Buffer size is set to max chunk size + 100 bytes for padding.
```
dirvine: Do we need to consider 100 bytes here for padding. It should be no more than 15 bytes I htink, but taken care of by the block_modes crate AFAIK.

src/encryption.rs
Yoga07: Ah! Missed it in the rush! Thanks! :)

src/encryption.rs
```diff
@@ -8,18 +8,46 @@
...
 
-pub fn encrypt(data: &[u8], key: &Key, iv: &Iv) -> Vec<u8> {
-    secretbox::seal(data, iv, key)
+pub fn encrypt<E>(data: &[u8], key: &Key, iv: &Iv) -> Result<Vec<u8>, SelfEncryptionError<E>>
+where
+    E: StorageError,
+{
+    let cipher = Aes128Cbc::new_var(key.0.as_ref(), iv.0.as_ref()).unwrap();
+    let pos = data.len();
+    let mut buffer = [0u8; BUFFER_SIZE];
```
madadam: This can overflow the stack, depending on the platform, compiler, etc... Better to use `Vec`.

src/encryption.rs
madadam: Or use [`encrypt_vec`](https://docs.rs/block-modes/0.3.3/block_modes/trait.BlockMode.html#method.encrypt_vec) (and `decrypt_vec`) and you don't have to worry about the padding at all.

src/encryption.rs
madadam: ...but as I wrote in the other comment, better to use `encrypt_vec`, then you don't have to worry about any of this.

tests/lib.rs
```diff
@@ -293,12 +293,9 @@ fn write_random_sizes_out_of_sequence_with_gaps_and_overlaps() {
 fn cross_platform_check() {
     #[rustfmt::skip]
     static EXPECTED_HASHES: [[u8; 32]; 3] = [
-        [98, 4, 121, 14, 80, 44, 103, 243, 182, 251, 248, 171, 11, 254, 131, 103, 156, 178, 138,
-         217, 103, 136, 72, 247, 125, 195, 212, 138, 115, 116, 227, 114],
-        [167, 132, 31, 67, 233, 190, 215, 173, 208, 106, 242, 71, 1, 114, 195, 50, 127, 158, 143,
-         187, 72, 11, 106, 107, 65, 219, 4, 243, 46, 43, 29, 116],
-        [133, 35, 122, 193, 107, 25, 207, 120, 214, 149, 158, 221, 143, 155, 90, 97, 5, 77, 94,
-         52, 235, 252, 6, 150, 80, 188, 143, 146, 116, 221, 107, 107],
+        [90, 123, 178, 77, 189, 56, 250, 228, 43, 186, 33, 61, 74, 91, 212, 16, 157, 230, 227, 31, 132, 167, 178, 127, 44, 33, 184, 3, 80, 29, 195, 41],
```
jeanphilippeD: Why did it change, are we not using the same algorithm?

Yoga07: Not exactly @jeanphilippeD :) We are updating it from rust_sodium's `secret_box::seal()` to `Cbc::encrypt_vec()` (which makes use of `Aes128`). Hence the change in hashes.

src/encryption.rs
```diff
@@ -19,21 +16,13 @@ type Aes128Cbc = Cbc<Aes128, Pkcs7>;
...
-    let pos = data.len();
-    let mut buffer = [0u8; BUFFER_SIZE];
-    buffer[..pos].copy_from_slice(data);
-    cipher
-        .encrypt(&mut buffer, pos)
-        .map(|res| res.to_vec())
-        .map_err(|_| SelfEncryptionError::Encryption)
+    let cipher = Aes128Cbc::new_var(key.0.as_ref(), iv.0.as_ref())
+        .map_err(|e| SelfEncryptionError::Cipher(format!("{:?}", e)))?;
+    Ok(cipher.encrypt_vec(data))
```
dirvine: Much nicer

PR #304, opened 2021-02-17
Delete trait for SEStorage

kanav99:
Implements delete trait for Storage which is needed for https://github.com/maidsafe/sn_client/pull/1377
This PR needs some tests.

### DIFF
--- a/examples/basic_encryptor.rs
+++ b/examples/basic_encryptor.rs
@@ -132,6 +132,13 @@ impl Storage for DiskBasedStorage {
             .map_err(From::from)
     }
 
+    async fn delete(&mut self, name: &[u8]) -> Result<(), SelfEncryptionError> {
+        let path = self.calculate_path(&name);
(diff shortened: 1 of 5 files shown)


lionel1704:
Looks good so far. But I think that the `delete()` function is missing from the `SelfEncryptor` struct. 
I imagine that the API would be something like:
```
let encryptor = SelfEncryptor::new(client, root_data_map);
encryptor.delete();
```

kanav99:
I thought that would go in the client, but I think this makes more sense. How should the state of the encryptor look like after delete? should we be able to write to it? or should it be closed?

lionel1704:
I think we can close the self encryptor after calling delete. It can consume the object instead of taking a reference.

lionel1704:
This looks good. It'd be nice to add some tests to this.

lionel1704 (approved):

PR #309, opened 2021-03-02
Stream Chunks on write and remove truncate

### LINKED ISSUE #258: Write chunks to network as encryption occurs
In the current implementation, when a file is self encrypted the chunks are held locally until `self_encryptor.close()` is called. When the SAFE Client Libraries consume these functions for Immutable Data the data is encrypted and stored locally and in the end all the chunks are sent to the network at once which increases the load on the network.

It would be nice to be able to start writing / streaming chunks to the network as they are being encrypted.

kanav99:
Fixes #258 
~~As of now, chunks are being stored twice, need to fix this.~~ This issue now fixed


src/self_encryptor.rs
```diff
@@ -227,65 +204,184 @@ where
...
+                state.chunks.clear();
+                state.sorted_map.clear();
+                state.sequencer.truncate(new_size as usize);
+                state.file_size = new_size;
                 return Ok(());
             }
 
-            if new_size >= state.file_size {
-                state.extend_sequencer_up_to(new_size);
+            {
```
maqi, line 248: better put this into `else` clause for clarification?

src/self_encryptor.rs
```diff
@@ -227,65 +204,184 @@ where
 
     /// Truncate the self_encryptor to the specified size (if extended, filled with `0u8`s).
     pub async fn truncate(&self, new_size: u64) -> Result<(), SelfEncryptionError> {
```
maqi, line 206: this function now becomes bit hard to follow the process flow.
I'd suggest have :
1, have some helper function to reduce the duplicated blocks
2, add a detailed comment (better with a table) to explain what happens under different branches

src/self_encryptor.rs
```diff
@@ -1483,6 +1715,81 @@ mod tests {
         Ok(())
     }
 
+    async fn general_truncate_test(
+        old_size: u32,
+        new_size: u32,
+        num_chunks: u32,
```
maqi, line 1721: this `num_chunks` is the `expected number of chunks after truncate` ?
better have a comment explain the general purpose of this function.

src/self_encryptor.rs
```diff
@@ -1483,6 +1715,81 @@ mod tests {
...
+            .expect("Third encryptor construction shouldn't fail.");
+        let fetched = se
+            .read(0, min_size as u64)
+            .await
+            .expect("Reading from encryptor shouldn't fail.");
+        assert_eq!(&fetched[..], &bytes[..(min_size) as usize]);
+        Ok(())
+    }
+    #[tokio::test]
+    async fn truncate_three_to_four_chunk() -> Result<(), SelfEncryptionError> {
```
maqi, line 1769: better have a comment explain the general setup and expected result of the test.
same comment to the below tests.

src/self_encryptor.rs
```diff
@@ -1483,6 +1715,81 @@ mod tests {
...
     }
 
+    async fn general_truncate_test(
+        old_size: u32,
+        new_size: u32,
+        num_chunks: u32,
+    ) -> Result<(), SelfEncryptionError> {
+        let bytes_len = old_size;
+        let mut rng = new_test_rng()?;
+        let bytes = random_bytes(&mut rng, bytes_len as usize);
```
maqi, line 1725: avoid using `keyword like` word as variable name.
better use something like `content`

src/self_encryptor.rs
```diff
@@ -1483,6 +1715,81 @@ mod tests {
...
+        };
+
+        let (data_map2, storage) = {
+            // Start with an existing data_map.
+            let se = SelfEncryptor::new(storage, data_map)
+                .expect("Second encryptor construction shouldn't fail.");
+            se.truncate(new_size as u64)
+                .await
+                .expect("Truncating encryptor shouldn't fail.");
+            se.close().await?
```
maqi, line 1744: if put a `check_file_size` before calling the `close`, shall it be expected to be `new_size` ?

src/self_encryptor.rs
```diff
@@ -1483,6 +1715,81 @@ mod tests {
...
+                .expect("Writing to encryptor shouldn't fail.");
+            check_file_size(&se, bytes_len as u64).await;
+            se.close().await?
+        };
+
+        let (data_map2, storage) = {
+            // Start with an existing data_map.
+            let se = SelfEncryptor::new(storage, data_map)
+                .expect("Second encryptor construction shouldn't fail.");
+            se.truncate(new_size as u64)
```
maqi, line 1741: if put a `check_file_size` before calling the `truncate`, shall it be expected to be `old_size` ?

maqi (commented):
Regarding the comment of the PR saying `chunks are being stored twice`, would like to know more about this.
If it means a concurrent store still happens when `close`, then it has to be fixed before the merge.

src/self_encryptor.rs
kanav99: Yes, I will add this check too

src/self_encryptor.rs
kanav99: Yes

src/self_encryptor.rs
kanav99: Yes I anticipated this problem. I will do this change

src/self_encryptor.rs
```diff
@@ -180,45 +180,22 @@ where
...
-        // - first two chunks if last chunks size has changed
-        // - chunks whose size is out of date
-        let (resized_start, resized_end) = {
-            let state = self.0.lock().await;
-            resized_chunks(state.map_size, state.file_size)
-        };
-
-        // end of range of possibly reusable chunks
-        let the_data_map = if resized_start == resized_end {
+        {
```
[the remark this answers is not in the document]
kanav99: Correct, added the change for the upcioming commit

tests/lib.rs
```diff
@@ -138,6 +138,89 @@ async fn new_read() -> Result<(), SelfEncryptionError> {
...
+    {
+        let slice_broken_data = &mut broken_data[..];
+        slice_broken_data.shuffle(&mut rng);
+    }
+
+    match broken_data.iter().filter(|&x| x.0 != last_piece).last() {
+        None => panic!("Should never occur. Error in test itself."),
+        Some(overlap) => {
+            let mut extra: Vec<u8> = overlap.1.to_vec();
+            extra.extend(random_bytes(&mut rng, 7usize)[..].iter().cloned());
```
maqi, line 175: define a const for this `majic` number

src/self_encryptor.rs
```diff
@@ -468,6 +375,110 @@ where
...
+
+    if new_size > old_size && old_size >= 3 * MIN_CHUNK_SIZE as u64 {
+        prepare_chunk_for_reading(Arc::clone(&state), 0).await?;
+        prepare_chunk_for_reading(Arc::clone(&state), 1).await?;
+        let mut state = state.lock().await;
+        state.chunks[0].flag_for_encryption();
+        state.chunks[1].flag_for_encryption();
+    }
+
+    if new_size < 3 * MIN_CHUNK_SIZE as u64 {
```
maqi, line 401: this check condition won't co-exist with the above one.
your or put this block `before` the above block (as it has a internal return), or put this as an else branch.

Also, better put a commet for each code block, as this function is somehow big.

.github/workflows/pr.yml
```diff
@@ -52,11 +51,11 @@ jobs:
     steps:
       - uses: actions/checkout@v2
         with:
-          fetch-depth: '0'
+          fetch-depth: "0"
```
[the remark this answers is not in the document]
kanav99: This was auto-prettified by prettier. Should I revert this change?

tests/lib.rs
kanav99: This code is just copied pasted from the previous function, just that we close after every write.

maqi (approved):

maqi (approved):

lionel1704 (approved):

PR #315, opened 2021-04-05
Refactor unsigned integer types

kanav99:
I am still unsure about the implications in the cross-platform scenario. According to me, there should be no problem as the encryption function calls and types are unchanged, so the output should not be different for any architecture. But @dirvine says that there are special ramifications we need to be sure of. But I went ahead and changed everything to `usize` and see what happens. If there are certain issues, we can specifically cast them there. But overall, the code looks much more cleaner with no explicit type casts everywhere.


lionel1704 (approved):

lionel1704:
The changes look good @kanav99 
Nice one! There's one more thing that's needed in this PR. Could you please add `BREAKING CHANGE` to any of the commit messages. So the message would be like:
```
refactor(module): what was done:

BREAKING CHANGE: what the breaking change is
```

kanav99:
Yes, the second commit has these breaking tag!

lionel1704:
Ah nice one! Sorry I didn't notice it

PR #332, opened 2021-08-24
SelfEncryption simplified

oetyng:
This is a simplification of the `SelfEncryption` library, which does the following:

- Removes the posix style API.
- Improves read performance.
- Improves write performance.
- Removes various inefficiencies and unnecessary code.
- Removes the reliance on a passed in storage, and lets the consumer deal with it entirely.
- Makes the internal hash size limit explicit to the library user.
- Clears up naming of public as well as private APIs.
- Makes the sensitive nature of the data map explicit by naming it secret key.
- Enforces min size of data passed in, for actual encryption to occur, as to not mislead users that anything was encrypted, when in fact it was not.


src/lib.rs
```diff
@@ -171,29 +90,271 @@
...
+        index_range: start_index..end_index,
+        start_pos: pos % get_chunk_size(file_size, start_index),
+    }
+}
+
+// ------------------------------------------------------------------------------
+//   ---------------------- Private methods -----------------------------------
+// ------------------------------------------------------------------------------
+
+/// Returns the chunk index range [start, end) that is overlapped by the byte range defined by `pos`
```
joshuef, line 210:
```suggestion
/// Returns the chunk index range [start, end] that is overlapped by the byte range defined by `pos`
```

src/lib.rs
```diff
@@ -171,29 +90,271 @@
...
+        || remainder >= MIN_CHUNK_SIZE
+        || position < file_size - remainder - MIN_CHUNK_SIZE
+    {
+        position / chunk_size
+    } else {
+        num_chunks - 1
+    }
+}
+
+// fn get_previous_chunk_index(file_size: usize, chunk_index: usize) -> usize {
```
joshuef, line 355: lets remove this if not needed

src/encryption.rs
```diff
@@ -1,27 +1,35 @@
...
-pub const IV_SIZE: usize = 16;
+pub(crate) const KEY_SIZE: usize = 16;
+pub(crate) const IV_SIZE: usize = 16;
+
+pub(crate) const HASH_SIZE: usize = XOR_NAME_LEN;
+pub(crate) const PAD_SIZE: usize = (HASH_SIZE * 3) - KEY_SIZE - IV_SIZE;
+
+pub(crate) struct Pad(pub [u8; PAD_SIZE]);
+pub(crate) struct Key(pub [u8; KEY_SIZE]);
+pub(crate) struct Iv(pub [u8; IV_SIZE]);
```
joshuef, line 25: What is `Iv` here? Perhaps we can rename/expand the acronym for clarity?

It's not immediately clear what `Pad` is all about here either tbh. Perhaps can you add a wee comment for each?

joshuef (changes requested):
Some wee small bits, and a couple of comments just foir clarity perhaps here.

Otherwise the commits will need a `BREAKING CHANGE: ` almost certainly :D 

but muccch easier to get what's going on here (and faster to boot!). nice onneeee 💥 🏄

src/lib.rs
oetyng:
```
A half-open interval includes only one of its endpoints, and is denoted by mixing the notations for open and closed intervals. For example, (0,1] means greater than 0 and less than or equal to 1, while [0,1) means greater than or equal to 0 and less than 1.
```

src/encryption.rs
oetyng: Ah, this is crypto lingo. Let's see if there's some commonly used variations, but at least some docs would be nice.

oetyng: IV: https://en.wikipedia.org/wiki/Initialization_vector
Padding: https://en.wikipedia.org/wiki/Padding_(cryptography)

oetyng: I think these are pretty standard names.

oetyng:
> will need a `BREAKING CHANGE: ` almost certainly :D
BREAKING CHANGE added :D 🏎️

src/encryption.rs
oetyng: It's funny, I have added the docs, but this comment doesn't show as outdated, and in the changes view, the docs doesn't show..

joshuef (approved):
good stuff. thumbs 👍

PR #336, opened 2021-09-02
Fix crates.io badge

atouchet:


connec (approved):
Thanks for the update!

PR #375, opened 2023-12-18
fix(decrypt): prevent extra clones while decrypting chunks

RolandSherwin:


src/decrypt.rs
```diff
@@ -11,42 +11,29 @@ use bytes::Bytes;
...
-                .jobs
                 .par_iter()
                 .map(|c| {
-                    Ok::<(usize, Bytes), Error>((
-                        c.index,
-                        decrypt_chunk(c.index, c.encrypted_content.clone(), c.src_hashes.as_ref())?,
-                    ))
+                    // we can pass &src_hashes since Rayon uses scopes under the hood which guarantees that threads are
+                    // joined before src_hashes goes out of scope
+                    let bytes = decrypt_chunk(c.index, c.content.clone(), &src_hashes)?;
```
joshuef, line 30: does this not work also? I dont see where `c` is being used after this?
```suggestion
                    let bytes = decrypt_chunk(c.index, c.content, &src_hashes)?;
```

joshuef (approved):

Issue #381, opened 2024-04-17
get_chunk_size() is returning incorrect file size when chunk index is > 1 and the file size is less than 3 * MAX_CHUNK_SIZE


traktion:
Thank you for contributing to the project!
We recommend you check out our ["Contributing to the SAFE Network"](https://github.com/maidsafe/QA/blob/master/CONTRIBUTING.md) guide if you haven't already.

**Describe the bug**

Originally found through integrating sn_client libraries into sn_httpd: https://github.com/maidsafe/safe_network/issues/1621#issue-2245520089

In short, when attempting to retrieve small parts of medium sized files (up to 1.5 MB), the relative_pos is incorrectly calculated for index 2+. This results in relative_pos being too low, causing an overlap with prior calls to `seek_info` when stepping through a file. 

**To Reproduce**
1. Call `seek_info` with a file smaller than 1.5 MB, using a small length such as 64 KB and increment the pos by 64 KB each time.
2. Once index 2 is referenced, the relative_pos will return duplicate bytes from the prior call.

**Expected behavior**
There should be no duplicate bytes returned from prior calls to seek_info().

**Screenshots**
This is a comparison between 2 files, one using a large length (512 KB) which is successful, vs one which used 128 KB and had duplicate bytes:

![Screenshot from 2024-04-16 09-53-55](https://github.com/maidsafe/self_encryption/assets/12068792/7a630e85-50ff-4d05-ac58-e41895a26dff)

**Desktop (please complete the followin
g information):**
 - OS: Linux

**Smartphone (please complete the following information):**
 - Device: Laptop

**Additional context**

traktion:
I have a PR to fix this issue, but I don't think I have permission to create a branch.

traktion:
PR here: https://github.com/maidsafe/self_encryption/pull/382

PR #382, opened 2024-04-17
Get chunk size fix

traktion:
- Updated the logic for calculating the chunk size when the chunk index is > 1 and the file size is less than 3 * MAX_CHUNK_SIZE (1.5 MB currently).
- The prior logic was incorrectly rounding, causing the size to be out by a small margin. This caused seek_info to return an incorrect relative_pos.
- Tested with 1 KB, 8 KB, 256 KB, 123456 bytes, 512 KB and 2048 KB over sn_httpd and confirmed all payloads were correct with Wireshark.

Link to issue:
https://github.com/maidsafe/self_encryption/issues/381

Link to issue:
https://github.com/maidsafe/safe_network/issues/1621

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -579,11 +579,7 @@ fn get_chunk_size(file_size: usize, chunk_index: usize) -> usize {
         return 0;
     }
     if file_size < 3 * MAX_CHUNK_SIZE {
-        if chunk_index < 2 {
-            return file_size / 3;
-        } else {
-            return file_size - (2 * (file_size / 3));
-        }
+        return file_size / 3;
     }
     let total_chunks = get_num_chunks(file_size);
     if chunk_index < total_chunks - 2 {
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -324,7 +324,6 @@ impl StreamSelfDecryptor {
     // If the file already exists, the content will be appended to the end of the file.
     fn append_to_file(&self, content: &Bytes) -> std::io::Result<()> {
         let mut file = OpenOptions::new()
-            .write(true)
             .append(true)
             .create(true)
             .open(&*self.file_path)?;
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -487,9 +487,16 @@ pub struct SeekInfo {
 /// with `relative_pos` from the `SeekInfo` instance, and the `len` to be read.
 pub fn seek_info(file_size: usize, pos: usize, len: usize) -> SeekInfo {
     let (start_index, end_index) = overlapped_chunks(file_size, pos, len);
+
+    let relative_pos = if start_index == 2 && file_size < 3 * MAX_CHUNK_SIZE {
+        pos - (2 * get_chunk_size(file_size, 0))
+    } else {
+        pos % get_chunk_size(file_size, start_index)
+    };
+
     SeekInfo {
         index_range: start_index..end_index,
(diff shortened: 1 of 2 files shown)


joshuef:
nice one @traktion thank you!

can you update for the clippy errors in ci [there](https://github.com/maidsafe/self_encryption/actions/runs/8722648758/job/23948078921?pr=382) that appears to b tripping up the other CI runs, I think 🙇

src/lib.rs
```diff
@@ -579,11 +579,7 @@ fn get_chunk_size(file_size: usize, chunk_index: usize) -> usize {
         return 0;
     }
     if file_size < 3 * MAX_CHUNK_SIZE {
-        if chunk_index < 2 {
-            return file_size / 3;
-        } else {
-            return file_size - (2 * (file_size / 3));
-        }
+        return file_size / 3;
```
maqi, line 582: I think the aim of the original code block is : 
always have chunks[2] to be the one holding extra bytes, when file_size is small AND non-dividable by 3.
say, if file_size is 1024, then chunk[0] & chunk[1] to be 341, chunk[2] to be 342

with the change, all three chunks will then be 341, which will cause a data loss ?

maqi (changes requested):

src/lib.rs
traktion: I can create a unit test with the above scenario and test it out. It's good to understand the context behind this.

From testing, it definitely wasn't working right and this seemed to fix it. However, it sounds like there could be the edge cases described.

traktion: Considering the above and looking at the code again, I think the bug is actually in the `seek_info` routine. I see there are a bunch of others that use `get_chunk_size` and they likely need the logic as you describe.

Given seeking is always from the start of the chunk and only the last chunk is oversized, we can use `get_chunk_size(file_size, 0)` in all cases.

traktion: I think there is also an issue with `overlapped_chunks` due to `get_chunk_index`, as when you have a large 3rd chunk, it calls `get_chunk_size(filesize, 0)` (like I thought may work above). However, if you have an oversized 3rd chunk, it returns 0 (instead of 2).

In turn, that means `seek_info` reverts to start_index == 0, which corrupts the relative_pos value.

maqi: Hi, @traktion ,

really appreciate your investigations and more tests is always appreciated.

meanwhile, I had a quick scan on self_encryption code, and doess see issues at
https://github.com/maidsafe/self_encryption/blob/master/src/lib.rs#L492
and 
https://github.com/maidsafe/self_encryption/blob/master/src/lib.rs#L640

L492 uses remainder when divided by chunk_size of starting chunk, which will be incorrect when:
a, the starting chunk is oversized chunk
b, the position falls into the edge of the oversized chunk

L640 bearing the similar issue that using chunk_size of the first chunk.

I will see if can get some fixes and tests to it on Monday when back.
And will be really appreciated and merge your tests in, if you come up more, or even with a solution to it.

Thank you very much.

maqi: great work !
really appreciate it.

maqi (approved):

traktion:
I'm not sure why the build is snagging on the above. It doesn't look like there should be any merge conflicts (no changes in months) and I've only _added_ code coverage. Please lmk if there is something I need to fix! Thanks!

PR #387, opened 2024-09-17
feat: default size back to 1mb

joshuef:


joshuef:
closed as #389 goes in

PR #388, opened 2024-09-25
feat: make chunk size user defined

joshuef:
BEAKING CHANGE: all APIs updated to have min/max_encryptrable bytes passed in by the user.

This allows for varying the use of the lib.


maqi (commented):
Just a general question that, to miniumn the impact of the change, why not make CHUNK_SIZE configurable via env ?
something like : 
```rust
lazy_static! {
    static ref MAX_CHUNK_SIZE: usize = std::env::var("MAX_CHUNK_SIZE")
            .unwrap_or("1048576".to_string())
            .parse::<usize>()
            .unwrap_or(1048576);
}
```

with that, the change can then be `non breaking` and be totally backward compatible.

On calling side, there will be no need to change any production code, just a `setup configuration` for each devnet.

b-zee:
> make CHUNK_SIZE configurable via env

Good idea, but it's possible to use without `lazy_static!` with `std::option_env!` (https://doc.rust-lang.org/std/macro.option_env.html)

maqi:
> Good idea, but it's possible to use without `lazy_static!` with `std::option_env!`

`lazy_static!` is still required to construct only once during run time ?

meanwhile, I do agree using `std::option_env` will be better to avoid pollution during run_time env

joshuef:
Okay as suggested, closing in favor of: https://github.com/maidsafe/self_encryption/pull/389

PR #391, opened 2024-10-15
feat: make chunk size user defined

joshuef:
BEAKING CHANGE: all APIs updated to have min/max_encryptrable bytes passed in by the user.

This allows for varying the use of the lib.

### DIFF
--- a/benches/lib.rs
+++ b/benches/lib.rs
@@ -51,6 +51,11 @@ use std::time::Duration;
 // https://bheisler.github.io/criterion.rs/book/analysis.html#measurement
 const SAMPLE_SIZE: usize = 20;
 
+/// The maximum size (before compression) of an individual chunk of a file, defined as 1024kiB.
+const MAX_CHUNK_SIZE: usize = 1024 * 1024;
+/// The minimum size (before compression) of an individual chunk of a file, defined as 1B.
+const MIN_CHUNK_SIZE: usize = 1;
+
 fn custom_criterion() -> Criterion {
     Criterion::default()
         .measurement_time(Duration::from_secs(40))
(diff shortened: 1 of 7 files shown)


src/lib.rs
```diff
@@ -411,16 +407,21 @@ pub fn decrypt_from_chunk_files(
 /// Encrypts a set of bytes and returns the encrypted data together with
 /// the data map that is derived from the input data, and is used to later decrypt the encrypted data.
 /// Returns an error if the size is too small for self-encryption.
-/// Only files larger than 3072 bytes (3 * MIN_CHUNK_SIZE) can be self-encrypted.
+/// Only files larger than 3072 bytes (3 * min_chunk_size) can be self-encrypted.
```
maqi, line 410: the number of 3072 shall be changed (same to the other place)

benches/lib.rs
```diff
@@ -51,6 +51,11 @@ use std::time::Duration;
 // https://bheisler.github.io/criterion.rs/book/analysis.html#measurement
 const SAMPLE_SIZE: usize = 20;
 
+/// The maximum size (before compression) of an individual chunk of a file, defined as 1024kiB.
+const MAX_CHUNK_SIZE: usize = 1024 * 1024;
```
maqi, line 55: better make these two as `consts for test usage only` ?
as I see they got used in couple of testing places

src/lib.rs
```diff
@@ -511,7 +512,12 @@ pub fn seek_info(file_size: usize, pos: usize, len: usize) -> SeekInfo {
...
 /// and `len`. Returns empty range if `file_size` is so small that there are no chunks.
-fn overlapped_chunks(file_size: usize, pos: usize, len: usize) -> (usize, usize) {
+fn overlapped_chunks(
+    file_size: usize,
+    pos: usize,
+    len: usize,
+    max_chunk_size: usize,
+) -> (usize, usize) {
     // FIX THIS SHOULD NOT BE ALLOWED
     if file_size < (3 * MIN_CHUNK_SIZE) || pos >= file_size || len == 0 {
```
maqi, line 522: there shall no longer be a const of `MIN_CHUNK_SIZE` in production code anymore ?

PR #392, opened 2024-10-30
Fix CBC padding issues in chunk decryption and remove index from EncryptedChunk 

dirvine:
# Fix CBC padding issues in chunk decryption

## Summary
This PR fixes issues with CBC padding errors during chunk decryption by improving how we handle chunk boundaries and ordering. The main changes focus on ensuring proper chunk handling during decryption operations, particularly during seek operations.

## Changes
- Remove parallel processing in decrypt to maintain proper chunk boundaries and ordering
- Improve chunk ordering logic using DataMap indices
- Add better error messages for missing chunks
- Fix seek_and_join test to properly handle chunk boundaries
- Fix overflow issues in seek_with_length_over_data_size test
- Add more defensive checks for chunk boundaries

## Testing
All tests are now passing, including:
- seek_and_join
- seek_with_length_over_data_size
- seek_over_chunk_limit

## Impact
This change improves reliability of chunk decryption without changing the underlying encryption format, maintaining backward compatibility with existing encrypted data.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ license = "GPL-3.0"
 name = "self_encryption"
 readme = "README.md"
 repository = "https://github.com/maidsafe/self_encryption"
-version = "0.30.0"
+version = "0.30.1"
 
 [dependencies]
 aes = "~0.8.1"
--- a/src/decrypt.rs
+++ b/src/decrypt.rs
@@ -8,63 +8,42 @@
 
 use crate::{encryption, get_pad_key_and_iv, xor, EncryptedChunk, Error, Result};
 use bytes::Bytes;
-use itertools::Itertools;
-use rayon::prelude::*;
 use std::io::Cursor;
 use xor_name::XorName;
 
 pub fn decrypt(src_hashes: Vec<XorName>, encrypted_chunks: &[&EncryptedChunk]) -> Result<Bytes> {
-    let num_chunks = encrypted_chunks.len();
-    let cpus = num_cpus::get();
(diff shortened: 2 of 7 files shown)


src/decrypt.rs
```diff
@@ -8,63 +8,42 @@
...
 use xor_name::XorName;
 
 pub fn decrypt(src_hashes: Vec<XorName>, encrypted_chunks: &[&EncryptedChunk]) -> Result<Bytes> {
-    let num_chunks = encrypted_chunks.len();
-    let cpus = num_cpus::get();
-    let batch_size = usize::max(1, (num_chunks as f64 / cpus as f64).ceil() as usize);
-
-    let raw_chunks: Vec<(usize, Bytes)> = encrypted_chunks
-        .chunks(batch_size)
-        .par_bridge()
```
b-zee, line 23: I am not sure we want to get rid of parallelization here. Has that been any issue recently?

dirvine: I am on the get it working 100% correct first, then speed up later, if that makes sense

Cargo.toml
```diff
@@ -8,7 +8,7 @@ license = "GPL-3.0"
 name = "self_encryption"
 readme = "README.md"
 repository = "https://github.com/maidsafe/self_encryption"
-version = "0.30.0"
+version = "0.30.1"
```
jacderida, line 11: Just a quick question. If this is a breaking change, do we want to bump to `0.31.0`?

dirvine: Ah it's not breaking this one.

PR #408, opened 2025-05-30
Remove unused tiny_keccak from dependencies

loziniak:


b-zee (commented):
Thanks for this one @loziniak, much appreciated.

I removed more dependencies in an additional commit.

PR #417, opened 2025-08-29
feat: add new streaming decryption without disk use

grumbach:


src/stream.rs
```diff
@@ -1,599 +1,422 @@
+// Copyright 2021 MaidSafe.net limited.
```
maqi, line 1: 2025

src/stream.rs
```diff
@@ -1,599 +1,422 @@
...
+    chunk_infos: Vec<ChunkInfo>,
+    src_hashes: Vec<XorName>,
+    get_chunk_parallel: F,
+    current_batch_start: usize,
+    current_batch_chunks: Vec<Bytes>,
+    current_batch_index: usize,
+    batch_size: usize,
 }
 
-impl StreamSelfEncryptor {
```
maqi, line 36: The PR just provides a non-disk-file dependent streaming decryptor.

Prefer leave StreamSelfEncryptor unless there is a new replacement presents,
to avoid the feature got `forgotten` .

maqi (changes requested):
The PR just provides a non-disk-file dependent streaming decryptor.
Prefer leave `StreamSelfEncryptor` unless there is a new replacement presents, to avoid the feature got `forgotten` . 

Also, the old `StreamSelfEncryptor ` avoids the `additional 3 chunks` that user needs to pay when use new datamap format.
Which could be used for some cost sensitive cases.

Hence, suggest to separate the `removal work` into an individual PR.

PR #418, opened 2025-09-01
In memory streaming decryptor

maqi:
The exisiting `StreamSelfDecryptor` and `StreamSelfEncryptor` are disk dependent,
which shall be replaced entirely by all in-memory ones.

However, as the old ones doesn't enforce the usage of 3 additional datamap chunks,
they may still need to be presebted in reposiotory till a final decision got made and an all-in-memory streaming encryptor got deployed.

This is a replacement work on top of PR https://github.com/maidsafe/self_encryption/pull/417
Close https://github.com/maidsafe/self_encryption/pull/417

* exisiting `StreamSelfDecryptor` and `StreamSelfEncryptor` moved into `stream_old` mod.
* new `streaming_decrypt` and `StreamingDecrypt` which allows an all-in-memory streaming decryption
* allows sequential or random access via the new all-in-memory streaming decryption


grumbach (approved):
LGTM!

PR #421, opened 2025-09-02
feat: implement all-in-memory StreamingEncrypt without disk dependent

maqi:
* Created an all-in-memory `StreamingEncrypt` without disk dependent, to be consistent with `StreamingDecrypt`
* Created correspondent tests, include encryption with content being pushed in out-of-order.
* Refactored `streaming_encrypt_from_file` helper function to utilise the new `StreamingEncrypt`
* Reorganized `StreamingDecrypt` into `stream_decrypt.rs`, and put `StreamingEncrypt` into `stream_decrypt.rs`

### DIFF
--- a/src/stream_encrypt.rs
+++ b/src/stream_encrypt.rs
@@ -1,0 +1,1238 @@
+// Copyright 2025 MaidSafe.net limited.
+//
+// This SAFE Network Software is licensed to you under The General Public License (GPL), version 3.
+// Unless required by applicable law or agreed to in writing, the SAFE Network Software distributed
+// under the GPL Licence is distributed on an "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+// KIND, either express or implied. Please review the Licences for the specific language governing
+// permissions and limitations relating to use of the SAFE Network Software.
+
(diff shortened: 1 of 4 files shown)


src/stream_encrypt.rs
```diff
@@ -0,0 +1,1238 @@
...
+    ///
+    /// // Process first chunk of data
+    /// let data1 = Bytes::from(vec![1u8; 5000]);
+    /// let result1 = encryptor.encrypt_part(0, data1)?;
+    /// assert!(result1.is_none()); // Not complete yet
+    ///
+    /// // Process second chunk of data
+    /// let data2 = Bytes::from(vec![2u8; 5000]);
+    /// let result2 = encryptor.encrypt_part(5000, data2)?;
+    /// assert!(result2.is_some()); // Complete! Returns DataMap
```
grumbach, line 342: This is kind of hard to use, I feel like this could be done in one function call without the whole indexing process which is very error prone.

grumbach (changes requested):
Nice one! Although this approach is quite hard to use. Why don't we simplify it to a single function that takes in a byte iterator and returns a chunk iterator?

grumbach:
Replaced by #422

PR #422, opened 2025-09-03
Improved in memory streaming encryptor

grumbach:
```rust
    let mut stream = stream_encrypt(data_size, data_iter)?;

    // Collect all chunks
    for chunk_result in stream.chunks() {
        let (xorname, chunk) = chunk_result?;
        // process chunks
    }

    // Get the datamap
    let datamap = stream.datamap()?;
```

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -96,7 +96,8 @@ mod encrypt;
 mod error;
 #[cfg(feature = "python")]
 mod python;
-mod stream;
+mod stream_decrypt;
+mod stream_encrypt;
 /// Contains old disk dependent streaming decryptor/encryptor,
 /// which using old DataMap format that doesn't enforce additional datamap chunks to be used.
 mod stream_old;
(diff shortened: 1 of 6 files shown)


maqi (changes requested):
Though this PR is on top of the PR [421](https://github.com/maidsafe/self_encryption/pull/421), the capability of out-of-order content push got removed (and the correspondent test as well.)  

The  `stream_encrypt_from_file` shall utilised the newly StreamEncrypt (or EncryptionStream) struct as well (like the PR 421 does).
So that we can avoid having too many different workflow path, which results in headache for future maintaince, like what we had in ant_cli/autonomi_client .

Regarding the `stream_encrypt` , I'd suggest to make it as a one-off help function as well (similar to stream_encrypt_from_file ):
```rust
fn stream_encrypt(data_size: usize, bytes_iterator: I, chunk_store: S) -> Result<DataMap>;
```
which we then can use it like this:
```rust
let datamap = stream_encrypt.datamap(data_size, bytes_iterator, chunk_store)?;
```

maqi (approved):
As discussed in the HO,
it is decided that:
* `out-of-order` content push is not mandatory for streaming_encrypt
* simpler API is preferred, unless there is critial user requirement to be asked for

Hence, got this PR approved to replace the [PR 421]( https://github.com/maidsafe/self_encryption/pull/421)

PR #425, opened 2025-09-08
fix: random streaming decryption handling different MAX_CHUNK_SIZE scheme

maqi:
`get_chunk_index()` made assumptions about uniform chunk sizes based on MAX_CHUNK_SIZE
For StreamingDecrypt, it shall use actual chunk sizes from the data map to calculate, to ensure consistent results regardless of the MAX_CHUNK_SIZE used during encryption.


grumbach (approved):
LGTM

PR #431, opened 2025-10-22
feat: only get datamaps from file

maqi:
1. Created a new public helper function: `streaming_encrypt_datamaps_from_file`
This function performs the core encryption logic but returns both the original and shrunk data maps without storing any chunks. It's memory-efficient and useful for scenarios where you need the data maps for analysis or want to handle chunk storage separately.
Key features:
* Returns a tuple (DataMap, DataMap) - original map and shrunk map
* Uses placeholder chunk store functions to avoid memory overhead
* Maintains the same encryption logic as the original function
* Includes comprehensive documentation with examples
2. Refactored the original `streaming_encrypt_from_file` function
The original function now uses a shared core implementation (streaming_encrypt_core) that can optionally store chunks. This eliminates code duplication while maintaining the same public API and functionality.
3. Added comprehensive unit tests

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -112,7 +112,7 @@ pub use self::{
     error::{Error, Result},
     stream_decrypt::{streaming_decrypt, DecryptionStream},
     stream_encrypt::{stream_encrypt, ChunkStream, EncryptionStream},
-    stream_file::{streaming_decrypt_from_storage, streaming_encrypt_from_file},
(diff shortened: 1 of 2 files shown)


dirvine (approved):
Nice work @maqi I do feel we need an API overhaul in this crate with perhaps a new cleaner API and deprecate old API over time.

PR #432, opened 2025-10-23
feat: env var of STREAM_DECRYPT_BATCH_SIZE

maqi:
make decryption batch size a configurable env var STREAM_DECRYPT_BATCH_SIZE, which defaults to 10 as current.
It's also public exposed for callers to use.


grumbach (approved):
LGTM

PR #442, opened 2026-05-12
feat: add per-stream decrypt batch sizing

mickvandijke:
## Summary
- add explicit per-stream decrypt batch-size APIs alongside the existing env-var default
- keep the legacy streaming_decrypt wrapper behavior unchanged
- bump self_encryption to 0.36.0 for the public API addition

## Verification
- cargo test

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 *When editing this file, please respect a line length of 100.*
 
+## [0.36.0] - 2026-05-13
+
+### Added
+- Added `streaming_decrypt_with_batch_size` and `DecryptionStream::new_with_batch_size` for
+  per-stream tuning of decrypt chunk fetch batches.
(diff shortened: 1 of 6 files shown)


dirvine (approved):
Approved. I did a local review and validation of this PR and did not find any blocking issues.

Non-blocking notes I recommend fixing:

- The new batch_size only affects sequential DecryptionStream iteration via fetch_next_batch. Random-access APIs such as get_range/range still fetch the full required chunk range in one call. That may be fine behaviorally, but the docs should clarify that this is iterator fetch batching only, or the batch-size behavior should be extended if callers are expected to rely on it for all streaming decrypt access paths.

- The PR adds tests for explicit batch sizing and explicit zero rejection, but not for the legacy STREAM_DECRYPT_BATCH_SIZE env fallback when the value is invalid or 0. Since that fallback behavior changed, it should have regression coverage if feasible despite the LazyLock isolation wrinkle.

These are not merge blockers from my review, but I suggest addressing them before merge or in a prompt follow-up.

jacderida (approved):

